PR #11, opened 2026-06-09
docs: use unified Hydro navbar and docs styling

shadaj:


docs/src/theme/Navbar/Content/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+  left: ReactNode;
+  right: ReactNode;
+}) {
+  return (
+    <div className="navbar__inner">
+      <div
+        className={clsx(
+          ThemeClassNames.layout.navbar.containerLeft,
+          'navbar__items',
+        )}>
```
akainth015, line 59: Is `clsx` genuinely useful here? A simple format string would do just as well, right? `${ThemeClassNames.layout.navbar.containerLeft} navbar__items`

docs/src/theme/Navbar/Content/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+import clsx from 'clsx';
+import {
+  useThemeConfig,
+  ErrorCauseBoundary,
+  ThemeClassNames,
+} from '@docusaurus/theme-common';
+import {
+  splitNavbarItems,
+  useNavbarMobileSidebar,
+} from '@docusaurus/theme-common/internal';
```
akainth015, line 11: Should we be using `@docusaurus/theme-common/internal`? It may be frustrating to update this as we pick up `docusaurus` updates.

docs/src/theme/Navbar/Content/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+        <ErrorCauseBoundary
+          key={i}
+          onError={(error) =>
+            new Error(
+              `A theme navbar item failed to render.
+Please double-check the following navbar item (themeConfig.navbar.items) of your Docusaurus config:
+${JSON.stringify(item, null, 2)}`,
+              {cause: error},
+            )
+          }>
```
akainth015, line 38: Is the navbar content static? In that case, does an error boundary complicate the tree unnecessarily?

akainth015 (approved):
I have a few hygiene / nit comments. They aren't blocking, though I think it's worth fixing them the next time they're touched.

docs/src/theme/Navbar/Content/index.tsx
shadaj: Yeah 🤷 probably not necessary, though we have the dep anyway 😆

docs/src/theme/Navbar/Content/index.tsx
shadaj: I think we kinda have to, because we're customizing the navbar by effectively patching it, indeed hopefully we don't run into compatibility issues.

docs/src/theme/Navbar/Content/index.tsx
shadaj: I think this is also just a result of the patching approach, this is the original code from Docusaurus. So I'll probably leave here for now but perhaps can revisit.

PR #15, opened 2026-06-10
feat(infinity-daemon): pass CLI-selected model to new sessions

shadaj:
When a user switched models in the CLI and then started a new session, the
selection was ignored: the picker only sent `SwitchModel` (which needs an
active session) and new sessions always booted with the daemon's default
model.

## Protocol (`infinity-protocol`)

* Add optional `model_id` to `ClientMessage::CreateSession` (`#[serde(default)]`
  for backward compatibility).

## CLI (`infinity-agent-cli`)

* Track the most recently picked model id in a `selected_model_id` local in
  `daemon_client.rs`; store it on every selection regardless of whether a
  session is active, and still forward `SwitchModel` when one is.
* Pass `selected_model_id` when sending `CreateSession`.
* Set `model_id: None` on the headless and ACP `CreateSession` constructions.

## Daemon (`infinity-daemon`)

* `client_handler` threads `model_id` through `CreateSession` (local branch
  into `create_session` and the remote-proxy forward).
* `SessionManager::create_session` accepts an optional `model_id`, validates
  it against `available_models` (returns an error for an unknown id instead of
  silently falling back), and threads it through `start_session` ->
  `start_agent_loop`.
* Add a shared `select_model` helper resolving an optional model id to a
  `ModelEntry` (default index 0), used by both `build_connected` and
  `start_agent_loop`. `Connected` now reports the chosen model's name and
  context window.

## UI (`infinity-ui`)

* Add the optional `model_id` field to the `CreateSession` TypeScript type.

### DIFF
--- a/crates/infinity-agent-cli/src/acp_server.rs
+++ b/crates/infinity-agent-cli/src/acp_server.rs
@@ -508,6 +508,7 @@ async fn run_session_connection(
             let msg = ClientMessage::CreateSession {
                 cwd,
                 location: None,
+                model_id: None,
             };
             send(&mut framed, &msg).await?;
             // Wait for Connected to get daemon ID, then send input.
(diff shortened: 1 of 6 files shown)


MingweiSamuel (approved):
LGTM

PR #18, opened 2026-06-10
refactor!: make model providers extensible via a dyn-compatible `ModelProvider` trait

shadaj:
## `infinity-agent-core`
* New `model_provider` module:
  * `ModelProvider` trait (via `async_trait`) exposing `list_models()` and `invoke_model(model_id, CompletionRequest)`, which returns the same `StreamingCompletionResponse` as rig's `CompletionModel::stream` with the provider-specific final response erased to `ProviderStreamingResponse` (carries token usage only).
  * Providers have no identity of their own — ids are assigned at registration by callers that manage multiple providers.
  * `erase_streaming_response()` helper and a `SingleModelProvider<M>` adapter wrapping any rig `CompletionModel` (used by tests).
* `run_completion`/`process_batch` now take a generic `P: ModelProvider + ?Sized` + model id instead of `Mdl: CompletionModel`, and no longer take `additional_request_params` / `model_id_override` / `max_output_tokens` — provider implementations handle those internally. Concrete callers (e.g. the Lambda) avoid `dyn` entirely and can specialize.

## `infinity-provider-bedrock` (new crate)
* `BedrockProvider` wraps the rig-bedrock client and handles all Bedrock-specific request parameters internally: adaptive-thinking config, per-model `additional_model_request_fields` (e.g. anthropic beta flags), and max output tokens.
* The two opus-4-6 configurations get distinct provider-scoped model ids (`...-v1` and `...-v1:1m`) that both map to the same underlying Bedrock model.

## `infinity-protocol`
* New `ModelRef { provider_id, model_id }` — models are now identified globally by provider id + model id.
* `ModelInfo` gains `provider_id`; `CreateSession`/`SwitchModel` carry `ModelRef` (**breaking**).

## `infinity-daemon`
* `model_picker.rs` replaced by `models.rs` with `ModelCatalog`: providers stored in a `HashMap` keyed by stable unique non-empty ids (asserted, since the empty id is the metadata serde sentinel); just Bedrock registered for now, but all code paths support multiple providers.
* Per-thread model selection: `ThreadInfo` gains a non-optional `selected_model: ModelRef`, assigned at thread creation (no parent fallback) and backfilled with the default model when loading metadata serialized before this change. `thread_worker` resolves its thread's model (and context window) from the catalog at startup, falling back to the global default with a warning if the stored model is gone.
* Fixes the bug where a session created with a specific model reverted to the default after a daemon restart — the selection is now persisted in thread metadata and re-resolved on every worker start.
* Removed the never-written `active_model_id`/`additional_request_params` RwLocks and the dead `Session::model_name`/`context_window` fields.

## `infinity-agent-lambda`
* Instantiates the concrete `BedrockProvider` (no `dyn`) and invokes it with the existing hardcoded model id.

## CLI / web
* CLI model picker now uses `infinity_protocol::ModelInfo` (no longer re-exports daemon types); selections are sent as `ModelRef`. Dropped the unused `rig-bedrock` dependency.
* `infinity-ui` `ModelInfo` type gains `provider_id`.

`./check.bash` passes except two pre-existing, unrelated `sandbox-local` environment-specific test failures.


crates/infinity-agent-core/src/batch_processor.rs
```diff
@@ -567,7 +565,24 @@ mod tests {
...
+    ) {
+        let (model, ctrl) = mock_model();
+        let entry = ModelEntry {
+            model_id: "mock".to_owned(),
+            display_name: "mock".to_owned(),
+            context_window: 0,
+            max_output_tokens: None,
+        };
+        (SingleModelProvider::new(entry, model), ctrl)
+    }
```
MingweiSamuel: Duplicated code?

crates/infinity-agent-core/src/event_processor.rs
```diff
@@ -2102,10 +2088,26 @@ mod tests {
...
+    ) {
+        let (model, ctrl) = mock_model();
+        let entry = ModelEntry {
+            model_id: "mock".to_owned(),
+            display_name: "mock".to_owned(),
+            context_window: 0,
+            max_output_tokens: None,
+        };
+        (SingleModelProvider::new(entry, model), ctrl)
+    }
```
MingweiSamuel: Duplicated code?

MingweiSamuel (approved):

crates/infinity-daemon/src/memory_store.rs
```diff
@@ -403,6 +438,29 @@ impl InMemoryConversationStore {
...
+    /// Get the model selected for this specific thread. Does NOT fall back to
+    /// the parent thread — every thread is assigned a model at creation time.
+    pub fn get_thread_model(&self, thread_id: &str) -> ModelRef {
+        self.ensure_thread_metadata_loaded(thread_id);
+        let threads = self.threads.lock().expect("bug: mutex poisoned");
+        threads
+            .get(thread_id)
+            .map(|t| t.selected_model.clone())
+            .expect("bug: thread metadata missing in get_thread_model")
+    }
```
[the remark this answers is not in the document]
shadaj: Nah, not an issue, we ensure thread loaded right above.

PR #20, opened 2026-06-12
test(infinity-agent-cli): TUI snapshot-test infrastructure with vt100 + alacritty, documenting ~25 viewport bugs

shadaj:
Refactor the terminal UI's I/O onto traits so it can be driven against
virtual terminals, and add a deterministic insta snapshot-test suite
(52 tests / 97 snapshots) that reproduces and documents existing TUI
bugs — no fixes included; buggy baselines are committed as-is and
annotated with greppable `// BUG:` comments (34) for a later fixing pass.

## Terminal I/O abstraction (`src/term_io.rs`)

* `TermOut` trait: ANSI byte sink (`Write`) + `size()`,
  `cursor_position()`, raw-mode toggles — everything the TUI queries
  from the terminal.
* `EventSource` trait: cancel-safe async `wait_for_event()` +
  non-blocking `try_read_event()` for crossterm input events.
* Production impls `CrosstermTerm` (stdout) and `CrosstermEvents` (16ms
  spawn_blocking poll, preserving spinner-animation wakeups; replaces
  `terminal::poll_crossterm_event`).

## Refactors (no behavior change)

* `InlineViewport<T: TermOut>` owns its terminal instead of grabbing
  `io::stdout()` ad hoc; size/cursor queries go through the trait.
* `terminal::run(term, events, ...)` is generic over both traits;
  `cleanup()` / `set_terminal_title()` take the terminal explicitly.
* `std::time::Instant` → `tokio::time::Instant` in spinner paths so
  paused-clock tests are deterministic.
* `MoveToNextLine` emits `CSI B` + CR instead of the equivalent but
  less-supported `CSI E` (vt100 lacks CNL; vt100 0.16 is blocked by
  ratatui's `unicode-width =0.2.0` pin).
* `install.rs` / `daemon_client.rs` updated to the new APIs.

## Test harness (`tests/common/mod.rs`)

* `Emulator` trait with two backends covering both real-world terminal
  behavior classes:
  * `Vt100Emulator` — truncates on resize, no reflow (xterm-like);
  * `AlacrittyEmulator` (via `alacritty_terminal`) — rewraps wrapped
    lines through scrollback and translates cursor + DECSC saved cursor
    on resize (alacritty/kitty/VTE-like), the class where inline-viewport
    resize bugs live. Wide-char spacer cells rendered correctly.
* `VirtualTerm` (TermOut → shared emulator) + `ScriptedEvents`
  (channel-fed EventSource).
* `TuiHarness`: spawns the real `terminal::run` loop with mock channels
  (`rig_mock::MockStreamingResponse` as the response type); helpers
  `display()`, `display_for_thread()`, `key()`, `type_str()`,
  `resize()`, `tick()`, `advance_and_redraw()`, `screen()` and
  `screen_with_scrollback()` (framed character grid + scrollback +
  cursor state + terminal title).
* Determinism: `#[tokio::test(start_paused = true)]`; `settle()` is a
  1ms sleep completing only via tokio auto-advance once the UI task has
  drained its queues; animations advanced explicitly with
  `tokio::time::advance`. Verified flake-free across repeated locked
  (INSTA_UPDATE=no) runs.

## Test suites

* `tui_snapshots.rs` (10, vt100): startup, slash autocomplete + Tab,
  thinking → streaming → done, tool call + result replacing the thinking
  display, resize mid-stream, help overlay, child-thread rows, message
  submission, session load, spinner animation over time.
* `tui_reflow_snapshots.rs` (14, alacritty): wrapped-scrollback reflow on
  narrow/wide resize, deep scrollback, vertical shrink/grow (incl. next
  to spinner-height changes and mid-stream partial lines), coalesced
  resizes, resizes with pickers open, widen mid-stream, print racing the
  Resize notification.
* `tui_viewport_snapshots.rs` (14, alacritty): viewport height changes
  next to scrollback (spinner/autocomplete/pickers/choice/thread rows,
  multiline input), multiline Info / OAuth / subscription events, tiny
  terminals, /help on narrow terminals, prints while picker open,
  submit mid-stream, compound grow/collapse.
* `tui_widechar_snapshots.rs` (5, alacritty): CJK/emoji display-width vs
  char-count issues in wrap_tail, the input box, and the status row.
* `tui_flow_snapshots.rs` (9, mixed): zellij-style session-replay burst
  (session switch + full history replayed undrained, spinner toggling
  between prints — renders clean, locked as regression guard),
  diff/empty tool results, child close_thread + subscription prefixes,
  multiline bracketed paste, quit picker via soft detach, lazy new
  session, queued choices + external completion, sessions updated while
  picker open, slash-command feedback.

## Bugs documented (34 `// BUG:` annotated assertions)

Each buggy baseline has a greppable `// BUG: <explanation>` comment above
its assertion. Highlights:

* **data loss**: widening mid-stream overwrites reflowed text; a print
  racing the Resize event is eaten by a stale scroll region (and its
  aftermath bakes a duplicate status row into scrollback)
* **vertical/horizontal resize corruption**: ghost spinners and stale or
  duplicated borders that subsequent prints scroll into permanent
  history; picker-resize artifacts; session-picker shrink merging
  leftovers into the status row
* **wide chars**: wrap_tail clips CJK/emoji tails (chars vs columns); emoji
  eat the following space in the input box; emoji unbalance status-row
  padding
* **text/layout**: multi-line Info loses newlines; OAuth URL staircases
  (LF without CR); tiny terminals lose the input row; /help corrupts
  scrollback on narrow terminals; status-row collision at narrow widths
* **stale state**: session change never redraws (hidden cursor, stale
  context %); "waiting for tool call result" spinner text persists after
  the result rendered; lazy new session keeps the old terminal title;
  sessions_updated never repaints an open session picker

Verified-correct behaviors (spinner grow/shrink, vertical grow pulling
history, prints while picker open, streaming CJK, replay burst, compound
collapse, etc.) are locked in as regression guards for the upcoming
fixes.

Deps: `vt100` + `alacritty_terminal` added to workspace dev-deps;
infinity-agent-cli dev-dependencies on insta/vt100/alacritty_terminal/
rig-mock and tokio `test-util`.

Co-authored-by: Infinity 🤖


crates/infinity-agent-cli/src/install.rs
```diff
@@ -97,6 +95,8 @@ async fn run_cargo_install(
     let status = format!("installing {crate_name}...");
     draw_spinner(viewport, &spinner_start, &status)?;
 
+    let mut events = CrosstermEvents;
+
```
MingweiSamuel, line 99: any particular reason for instantiating this rather than just `CrosstermEvents.wait_for_event()`

MingweiSamuel (approved):
Nice! Looking forward to fixes

crates/infinity-agent-cli/src/install.rs
shadaj: Little bit of future proofing to allow testing this with sim backend.

PR #21, opened 2026-06-12
fix(infinity-agent-cli): fix all 34 snapshot-documented TUI bugs (reflow, wide chars, spinner state, tiny terminals)

shadaj:
Fixes every `// BUG:` annotated in the snapshot suite from the previous
commit, healing all baselines (53 tests, +1 new). Round-trip discipline got
*stricter*: the cursor is now queried exactly once per resize and never
otherwise (the old code queried on every viewport height change). All
rendering remains diff-based — no new flicker — and redraws never emit LF,
so they can't push the screen.

## Inline viewport: reflow-safe anchoring (`inline_viewport.rs`)

The root cause of the ~17 resize-corruption bugs: reflowing terminals
(kitty/alacritty/VTE) translate the **live** cursor through a resize reflow
but merely *clamp* the DECSC saved cursor that the viewport anchored
everything on. The viewport now re-derives its anchor from the live cursor:

* hidden cursors are *parked on the anchor*, so the terminal itself tracks
  the anchor through reflows for free;
* with the cursor shown in the input box, the anchor is reconstructed by
  computing how many rows the previously drawn viewport rows rewrap to at
  the new width (occupied lengths captured from the last frame, matching
  alacritty's cell-emptiness rules);
* non-reflowing terminals (xterm-class) are detected by the queried cursor
  matching its old position exactly, keeping the old anchor; the ambiguous
  width-grow case resolves in favor of reflow (cosmetic gap on xterm vs.
  data loss on kitty).
* The anchor's mid-line column (streaming) is recomputed for any width from
  a tracked logical-line length: `print_above` buffers its output and parses
  it (`OutputTracker`: CR/LF/tabs, SGR, display widths, auto-wrap), panicking
  on any escape sequence it doesn't know how to track rather than silently
  desynchronizing.
* `print_above` checks `size()` (cheap ioctl, no round trip) on every print,
  repairing the anchor before printing — fixes prints racing the Resize
  event eating lines or corrupting scrollback.
* After a resize the viewport stays pinned to the bottom: a later reflow can
  then push at most *blank* gap rows into scrollback — never stale viewport
  rows — and prints consume the gap (the anchor advances into it) before any
  scrolling resumes. The new `repeated_resizes_do_not_accumulate` test
  asserts both properties through a double width-shrink over wrapped
  scrollback and vertical shrink/grow cycles. (The alternative — drawing
  tight under the anchor after resizes, as the old code did — was probed and
  rejected: it leaks stale border rows into scrollback during resize storms.)
* Degenerate scroll regions (`CSI 1;1r` on tiny terminals) are skipped and
  followed by a repaint instead of being silently ignored by the terminal.
* Viewport height is clamped to rows−1 (the anchor needs a row above).

## Terminal UI logic (`terminal.rs`)

* `ToolResult` clears "waiting for tool call result" and switches the
  spinner once all pending root tool calls (counted) have resolved.
* Session-change and sessions-updated arms now redraw (status/% context,
  cursor restore, live session-picker list refresh).
* Switching to a lazily-created session (and `/archive`) resets the terminal
  title left over from the previous session.
* Multi-line `Info` and OAuth messages print line-by-line (ratatui `Line`
  drops `\n`; raw LF staircases in raw mode).
* `/help` box clamps to the terminal width and truncates rows instead of
  wrapping into scroll-region corruption.
* Status row measures display width (emoji/CJK), keeps the right side whole,
  and truncates the left with an ellipsis and a guaranteed 2-space gap.
* `wrap_tail` (thread rows, thinking text) trims by display width so the
  freshest CJK/emoji text stays visible.
* Tiny terminals shed gap → spinner → border rows before squeezing the
  input/status rows.

## Text input (`text_input.rs`)

Width-aware wrapping, cursor math, and rendering: wide chars occupy two
columns (no more swallowed characters after emoji), cursor lands on the
correct column.


crates/infinity-agent-cli/tests/tui_reflow_snapshots.rs
```diff
@@ -252,12 +233,71 @@ async fn vertical_shrink_mid_stream_partial_line() {
     });
     h.display(Evt::ResponseDone(None));
     h.settle().await;
-    // BUG: the stream continuation is printed on top of a stale border
-    // fragment ("──…── and then keeps going afterwards" merged on one row),
-    // with a ghost spinner row above the viewport.
```
MingweiSamuel, line 257: Might be good to keep the comments around for visibility, just re-prefixed

MingweiSamuel (approved):

MingweiSamuel:
Also mentioned that it would be good to have the ANSI codes in the snapshots

PR #22, opened 2026-06-12
fix(infinity-agent-cli): context usage resets to zero after session replay

shadaj:
## Root cause

When loading a session, the CLI sets the context counter from the
`Connected` message (`total_tokens_used` from session info). But after
forwarding the replayed history, `daemon_client` appended a synthesized
`ResponseDone(Some(DaemonTokenUsage(None)))` marker, and the terminal's
`ResponseDone` handler did:

```rust
if let Some(r) = r {
    total_tokens_used = r.token_usage().map_or(0, ...);
}
```

Since the marker carries no usage, `map_or(0, ...)` reset the counter to
zero right after the replay. (Not a daemon or Bedrock bug — though the
same pattern existed daemon-side, see below.)

## Changes

* `terminal.rs`: only update `total_tokens_used` when the response
  actually reports usage; a usage-less `ResponseDone` (post-replay
  marker, or a provider that omits usage metadata) keeps the last known
  value instead of zeroing it.
* `daemon_client.rs`: send the end-of-replay marker as
  `ResponseDone(None)` instead of wrapping an empty usage, making the
  "no usage info" intent explicit.
* `infinity-daemon/thread_worker.rs` (defensive, same bug class): the
  display-event forwarder no longer persists `total_tokens_used = 0`
  when a response has no usage metadata (e.g. a Bedrock stream that ends
  without a usage `Metadata` event); `last_updated` is still refreshed.

## Tests

* New snapshot test `replay_keeps_context_usage` modeling the real load
  flow: `SessionChanged` with 42k tokens → replayed history →
  `ResponseDone(None)` → `ResponseDone` with `usage: None`. Status bar
  must still show `42% context used` (previously showed `0%`).

### DIFF
--- a/crates/infinity-agent-cli/src/daemon_client.rs
+++ b/crates/infinity-agent-cli/src/daemon_client.rs
@@ -543,7 +543,10 @@ async fn run_client(
                                     let _ = display_tx.send(evt);
                                 }
                             }
-                            let _ = display_tx.send((None, DisplayEvent::ResponseDone(Some(DaemonTokenUsage(None)))));
+                            // End-of-replay marker: closes any open stream/spinner
+                            // state. Deliberately carries no usage info so it does
+                            // not clobber the total from the Connected message.
+                            let _ = display_tx.send((None, DisplayEvent::ResponseDone(None)));
                             for m in pending_choices {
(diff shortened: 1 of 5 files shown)


MingweiSamuel (approved):
Very nice

PR #24, opened 2026-06-12
fix(infinity-daemon): persist in-flight tool results when a session is stopped

shadaj:
Previously, stopping an agent (quit → "stop agent" → `ShutdownSession`) while
the model was processing a freshly-arrived tool result lost that result:
`session_wrapper` broke out of its select on `shutdown_rx` and **dropped** the
`agent_loop` future. That closed each thread worker's input channel, and the
worker's `rx.recv()` arm responded to the closed channel with a bare `return`,
dropping the in-flight completion future. Since pending history items (the
tool result, plus any partial assistant output) are only synced to the
conversation store at the end of a completion, the store was left ending at
the tool call — on reload the conversation showed the call but no result.

* `thread_worker`: when the input channel closes mid-completion, reuse the
  existing interrupt mechanism — send `completion_cancel_tx` and await the
  completion future. The cancellation path strips trailing reasoning and then
  syncs all pending history items to the store, so nothing is dropped except
  incomplete thinking.
* `agent_loop`: accept a shutdown signal and keep each worker's `JoinHandle`.
  On shutdown (or rx close), drop all worker channels — triggering each
  worker's graceful interrupt path — and join every worker before returning,
  so the session is never torn down underneath them.
* `session_wrapper`: on `shutdown_rx`, signal the agent loop and await
  `agent_fut` to completion (instead of dropping it) before killing RAP
  servers and marking the session shut down. The idle-exit path now winds the
  agent loop down the same way. The select is `biased` toward `agent_fut` so
  a self-completed loop is always consumed by its own arm, making the
  `.await` in the other arms safe. The old "drain idle pings" hack is gone
  since agent_loop now joins workers itself.
* Added regression test `shutdown_persists_in_flight_tool_result` covering
  the exact race: tool result arrives, model request is in flight, shutdown
  fires — the tool result must be in the conversation store afterwards.
  (Verified the test fails against the old code.)

### DIFF
--- a/crates/infinity-daemon/src/session/agent_loop.rs
+++ b/crates/infinity-daemon/src/session/agent_loop.rs
@@ -14,6 +14,7 @@ use crate::session::ActiveThreads;
 struct WorkerChannels {
     input_tx: mpsc::UnboundedSender<(infinity_agent_core::message::InputMessage, String)>,
     subscribe_tx: mpsc::UnboundedSender<SubscribeRequest>,
+    handle: tokio::task::JoinHandle<()>,
 }
 
 #[expect(
@@ -34,10 +35,20 @@ pub async fn agent_loop(
     subscriber_map: SubscriberMap,
     active_threads: ActiveThreads,
     idle_tx: mpsc::UnboundedSender<()>,
+    shutdown: tokio_util::sync::CancellationToken,
 ) {
     let mut workers: HashMap<String, WorkerChannels> = HashMap::new();
 
-    while let Some(msg) = rx.recv().await {
+    loop {
+        let msg = tokio::select! {
+            biased;
+            _ = shutdown.cancelled() => {
+                tracing::info!("Agent loop for session {} received shutdown signal", session_id);
+                None
+            }
+            msg = rx.recv() => msg,
+        };
+        let Some(msg) = msg else { break };
         let thread_id = match &msg {
             AgentMessage::Input(input, _) => input.group_id.clone(),
             AgentMessage::Subscribe { thread_id, .. } => thread_id.clone(),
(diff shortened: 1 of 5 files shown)


crates/infinity-daemon/src/session/agent_loop.rs
```diff
@@ -357,6 +518,7 @@ mod tests {
         let sender = InMemoryMessageSender::new(input_tx);
         let subscriber_map: SubscriberMap = Arc::new(std::sync::Mutex::new(HashMap::new()));
         let active_threads = Arc::new(std::sync::Mutex::new(HashSet::new()));
+        let (shutdown_tx, shutdown_rx) = tokio::sync::oneshot::channel();
```
MingweiSamuel: https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html may be better for this

MingweiSamuel: but I guess this is just a test utility anyway?

crates/infinity-daemon/src/session/mod.rs
```diff
@@ -704,6 +704,11 @@ fn spawn_agent_loop(
 
     let subscriber_map = Arc::new(std::sync::Mutex::new(HashMap::new()));
 
+    // Internal channel used by session_wrapper to tell the agent loop to wind
+    // down its thread workers (interrupting in-flight completions so they
+    // flush pending history) before the session is torn down.
+    let (worker_shutdown_tx, worker_shutdown_rx) = oneshot::channel();
```
MingweiSamuel: https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html may be better

crates/infinity-daemon/src/session/agent_loop.rs
```diff
@@ -34,10 +35,19 @@ pub async fn agent_loop(
...
     let mut workers: HashMap<String, WorkerChannels> = HashMap::new();
 
-    while let Some(msg) = rx.recv().await {
+    loop {
+        let msg = tokio::select! {
+            msg = rx.recv() => msg,
+            _ = &mut shutdown_rx => {
+                tracing::info!("Agent loop for session {} received shutdown signal", session_id);
+                None
+            }
```
MingweiSamuel, line 48: Should this be biased too?

MingweiSamuel (approved):

PR #30, opened 2026-06-16
fix(sandbox-local): pass `--ignore-working-copy` to jj cleanup commands

shadaj:
When a parent thread continues working after a child is done, the child's
jj workspace becomes stale. On `close_thread`, `jj workspace forget` and
`jj abandon` fail with "working copy is stale", leaving the child's
workspace and commit dangling in jj.

* Add `--ignore-working-copy` to `workspace forget` and `abandon` in
  `cleanup_sandbox_permanently` (async close_thread path)
* Add `--ignore-working-copy` to the same commands in `Drop for LocalBackend`
  (server shutdown sync path)
* Add `--ignore-working-copy` to `jj_bookmark_is_empty` (both async and sync)
* Add `--ignore-working-copy` to `jj git export` from the orig dir in
  `push_sandbox`
* Add regression test reproducing the production scenario

### DIFF
--- a/crates/sandbox-local/src/backend.rs
+++ b/crates/sandbox-local/src/backend.rs
@@ -144,7 +144,15 @@ impl LocalBackend {
 /// Check if a jj bookmark's commit is empty (no changes) using a synchronous command.
 fn jj_bookmark_is_empty(dir: &Path, bookmark: &str) -> bool {
     Command::new("jj")
-        .args(["log", "--no-graph", "-r", bookmark, "-T", "empty"])
+        .args([
+            "--ignore-working-copy",
+            "log",
+            "--no-graph",
+            "-r",
+            bookmark,
+            "-T",
+            "empty",
+        ])
         .current_dir(dir)
         .stdout(Stdio::piped())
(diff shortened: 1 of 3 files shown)


crates/sandbox-local/src/backend.rs
```diff
@@ -497,7 +505,7 @@ impl SandboxBackend for LocalBackend {
                         .remote_uri,
                 );
                 if orig.join(".git").exists()
-                    && let Err(e) = run_jj(&orig, &["git", "export"]).await
+                    && let Err(e) = run_jj(&orig, &["--ignore-working-copy", "git", "export"]).await
```
MingweiSamuel, line 508: this one seems a little suspicious, maybe it should update the workspace instead

MingweiSamuel (approved):

akainth015 (approved):

crates/sandbox-local/src/backend.rs
akainth015: Is running `jj git export` at this particular point (after the child thread is done working) even useful? I'd guess not...

PR #33, opened 2026-06-17
fix(infinity-daemon): spawned threads inherit parent's model instead of default

### LINKED ISSUE #32: Threads spawned using `spawn_thread` use default model instead of parent model
The child thread is inserted with:
```rust
selected_model: self.default_model.clone(),
```

but it should be inserted with the model that the parent is using.

shadaj:
Fixes #32. When `spawn_thread` created a child thread in the
`InMemoryConversationStore`, it assigned `self.default_model` as the
child's `selected_model`. This caused child threads to always use the
global default model regardless of what model the parent thread was
configured to use.

The fix reads the parent thread's `selected_model` and passes it to the
newly created child `ThreadInfo`.

* `crates/infinity-daemon/src/memory_store.rs`: In `spawn_thread`, read
  `parent_info.selected_model` from the parent's `ThreadInfo` and use it
  for the child instead of `self.default_model`.
* `crates/infinity-daemon/src/session/agent_loop.rs`: Add regression test
  `spawned_thread_inherits_parent_model` that sets up a two-provider
  catalog, assigns the non-default model to the root thread, triggers a
  `spawn_thread` tool call, and verifies the child thread's model request
  goes to the parent's provider (not the default).

### DIFF
--- a/crates/infinity-daemon/src/memory_store.rs
+++ b/crates/infinity-daemon/src/memory_store.rs
@@ -832,16 +832,20 @@ impl ConversationStore for InMemoryConversationStore {
         let new_id = uuid::Uuid::new_v4().to_string();
         let spawn_message_order;
         let root;
+        let parent_model;
         {
             let threads = self.threads.lock().expect("bug: mutex poisoned");
             let msgs = self.messages.lock().expect("bug: mutex poisoned");
             spawn_message_order = spawn_order_override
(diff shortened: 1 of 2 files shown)


MingweiSamuel (approved):
Seems good. Would be interesting to let the parent agent choose in the future (particularly if its a no-context child, once we also have that)

PR #39, opened 2026-06-18
fix(infinity-agent-core): compaction inside child thread no longer panics on indexing

### LINKED ISSUE #31: Compaction thread spawned _inside_ a child thread leads to improper indexing
Error of the form:
```
ERROR rap_protocol: task 'thread_worker' panicked: range end index 60 out of range for slice of length 38
```

shadaj:
When compaction triggered inside a child thread, `safe_spawn_point()` returned
an in-memory index that included ancestor messages prepended to the history.
This index was used as `spawn_order_override` for the compaction grandchild,
but the child's actual store only contained its own messages — causing a panic:
"range end index X out of range for slice of length Y".

The fix:
- `load_history_with_ancestors` now returns `ancestor_prefix_len` as a third
  tuple element (how many messages at the front come from ancestor threads).
- `HistoryManager` stores this in a `Cell<usize>` field.
- `safe_spawn_point()` subtracts `ancestor_prefix_len` so the returned index
  is relative to the thread's own store.
- `apply_compaction()` adds `ancestor_prefix_len` to the split position (since
  ancestors occupy the beginning of in-memory history) and resets it to 0 after
  compaction (ancestors are consumed into the summary).

A regression test (`compaction_inside_child_thread_does_not_panic`) reproduces
the exact panic from issue #31 and uses insta snapshots to verify both the
compaction child's inherited history and the post-compaction history.

Fixes #31 

### DIFF
--- a/crates/infinity-agent-core/src/event_processor.rs
+++ b/crates/infinity-agent-core/src/event_processor.rs
@@ -1,5 +1,5 @@
 use std::{
-    cell::RefCell,
+    cell::{Cell, RefCell},
     collections::{HashMap, HashSet},
     time::Duration,
 };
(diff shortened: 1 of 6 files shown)


MingweiSamuel (approved):
Seems like we should really add better abstractions over the history tree? Suspicious that there are so many different indices

shadaj:
> Seems like we should really add better abstractions over the history tree? Suspicious that there are so many different indices

Yeah, we are basically maintaining an externally-stored (in DSQL, for Lambda backend) skip-list with arbitrary forking. Hence lots of complexities creeping in that we should properly abstract away.

PR #40, opened 2026-06-18
fix: subscription events inherit pretty tool call display_as

shadaj:
In both the live processing path (batch_processor.rs) and the history
replay path (display.rs), subscription events now use the `display_as`
string from the originating `InfinityMessage::ToolCall` when available,
falling back to `name(args)` only when no pretty display was computed.

The history lookup iterates in reverse since the subscription tool call
is typically near the end of the conversation.

Added a unit test (`subscription_event_uses_display_as`) that seeds
history with a tool call + result (subscription started), then sends a
synthetic subscription event and asserts the display uses the pretty
`display_as` string rather than raw `name(args)`.

### DIFF
--- a/crates/infinity-agent-core/src/batch_processor.rs
+++ b/crates/infinity-agent-core/src/batch_processor.rs
@@ -12,11 +12,13 @@ use std::pin::Pin;
 
 use futures_util::StreamExt;
 use rig::completion::{GetTokenUsage, ToolDefinition};
-use rig::message::{AssistantContent, Message, ToolResultContent, UserContent};
+use rig::message::{ToolResultContent, UserContent};
 use tokio::sync::{mpsc, oneshot};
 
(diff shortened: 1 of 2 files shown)


crates/infinity-agent-core/src/batch_processor.rs
```diff
@@ -131,20 +133,17 @@ where
                 if let InputMessageContent::User(UserContent::ToolResult(res)) = &input_msg.content
                     && let ToolResultContent::Text(text) = res.content.first()
                 {
-                    let orig_call = current_history.get_history().into_iter().find(|h| {
-                        if let Message::Assistant { content, .. } = h
-                            && let AssistantContent::ToolCall(c) = content.first()
+                    let orig_call = current_history.history.borrow().iter().rev().find_map(|m| {
```
shadaj, line 136: This might not be correct if the tool call is from before a compaction point?

PR #49, opened 2026-07-02
fix(infinity-web): update provider name when loading a thread

shadaj:
The web UI's `Connected` message handler was setting `modelName` and
`contextWindow` but not `providerName`. The daemon already sends
`provider_id` in the `Connected` message, but the frontend wasn't
reading it.

* Added `provider_id: string` to the `Connected` type in `infinity-ui/src/types.ts`
* Added `setProviderName(p.provider_id)` in the `Connected` handler in `App.tsx`

### DIFF
--- a/infinity-web/src/App.tsx
+++ b/infinity-web/src/App.tsx
@@ -230,10 +230,12 @@ export function App() {
               context_window: number;
               title: string | null;
               total_tokens_used: number;
+              provider_id: string;
             }>(m);
             sessionRef.current = p.session_id;
             threadRef.current = p.thread_id;
             setModelName(p.model_name);
+            setProviderName(p.provider_id);
             setContextWindow(p.context_window);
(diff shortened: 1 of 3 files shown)


infinity-web/src/App.tsx
MingweiSamuel, line 900:
```suggestion
          {providerName && `${providerName}:`}
          {modelName}
```
nit: space no longer needed now that template has whitespace?

MingweiSamuel, line 901:
```suggestion
          {modelName && " \u00b7"}
```
(same here I guess)

MingweiSamuel (approved):

infinity-web/src/App.tsx
shadaj: Ah, JSX doesn't introduce whitespace between such adjacent blocks.

PR #50, opened 2026-07-02
perf(infinity-agent-core): coalesce consecutive streamed text chunks

shadaj:
When a model provider streams an assistant response as multiple text
chunks, each chunk was persisted as its own message row, inflating disk
usage.

* `handle_completion` now merges an incoming text chunk into the last
  pending item when that item is already an assistant text message,
  concatenating the new text instead of appending a new pending item.
* Added `try_merge_pending_text`, which also keeps the in-memory history
  entry in sync with the merged text. Because `sync` drains
  `pending_items`, merging only happens across not-yet-persisted chunks,
  so an already-persisted text message is never mutated.
* Added tests covering coalescing of consecutive text chunks and
  verifying that a non-text item (e.g. a tool call) between text chunks
  breaks the run so they are not merged.

### DIFF
--- a/crates/infinity-agent-core/src/event_processor.rs
+++ b/crates/infinity-agent-core/src/event_processor.rs
@@ -102,6 +102,7 @@ pub enum CompletionEvent<R> {
 
 // ── HistoryManager (unchanged from before) ──
 
+#[derive(Serialize, Clone)]
 pub struct PendingItem {
     message: InfinityMessage,
     message_id: String,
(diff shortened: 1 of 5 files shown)


crates/infinity-agent-core/src/event_processor.rs
```diff
@@ -334,6 +343,38 @@ impl<C: ConversationStore, S: StateStore> HistoryManager<C, S> {
...
+        // The last pending item always corresponds to the last history entry
+        // (both are pushed/popped together while an item is pending), so keep
+        // the in-memory history in sync with the merged text.
+        if let Some(InfinityMessage::Assistant {
+            content: AssistantContent::Text(hist_text),
+        }) = self.history.borrow_mut().last_mut()
+        {
+            hist_text.text.push_str(text);
+        }
+        true
```
MingweiSamuel: Should we panic here if not? `let ... else { panic!("pending_items and history out of sync") };`?

MingweiSamuel (approved):

crates/infinity-agent-core/src/event_processor.rs
shadaj: Good call — since the last pending item always corresponds to the last history entry, a mismatch is a logical bug. Changed the `if let` to a `let ... else { panic!("bug: pending_items and history out of sync") }`.

PR #55, opened 2026-07-03
test: add deterministic e2e tests for TUI↔daemon and web UI (Playwright)

shadaj:
## Daemon: injectable providers & configurable paths

* `SessionManager::with_providers(SessionManagerConfig, providers, processes)` —
  generic constructor taking explicit `(provider_id, provider)` pairs;
  `SessionManager::new` keeps the production defaults (spawn providers from
  `~/.infinity/providers.json`, user RAP config at `~/.infinity/rap.json`).
* `SessionManagerConfig { state_dir, callback_url, user_rap_config, id_source }`
  reduces hardcoded paths: `user_rap_config: None` makes sessions hermetic.
* `rap_callback::serve_callbacks(listener, manager)` — callback accept loop for a
  pre-built manager; `start_callback_server` now composes it.
* `ws_handler::serve(listener, mgr)` — reusable HTTP/WS accept loop (used by
  `run_daemon` and by tests binding an OS-assigned port).
* `boot_rap_servers` takes the user-level config path as a parameter.

## Deterministic ids (`infinity_daemon::ids::IdSource`)

User-visible ids (session ids from `create_session`, thread ids from the
conversation store) now come from an injectable `IdSource`. Production uses
random v4 UUIDs; tests use `IdSource::sequential()`
(`00000000-0000-4000-8000-000000000001`, …) so ids rendered in UIs are stable
across runs — no snapshot redaction or screenshot masking needed.

## TUI ↔ daemon e2e (`crates/infinity-agent-cli/tests/e2e_daemon_tui.rs`)

* `daemon_client` moved from the binary into the lib; `run_client` is now public
  and generic over `TermOut`/`EventSource` + explicit cwd (crossterm /
  `current_dir()` defaults stay in `run_with_daemon`/`run_in_memory`).
* Single-process test on a paused-clock current-thread runtime: real
  `SessionManager` (mock rig provider) + `handle_client_channels` + real
  `run_client` rendered into a vt100 virtual terminal. Covers lazy session
  creation from first input, streamed chunks, `set_title` tool round-trip
  (terminal title), multi-turn, quit-picker-when-busy, and clean Ctrl+C
  shutdown, with 6 insta screen snapshots (fully deterministic, no filters).

## Web UI e2e (`crates/infinity-daemon/tests/web_e2e.rs`, feature `e2e-web`)

* New `e2e-web` feature = `bundled-web` + optional `playwright-rs` dep (optional
  regular dep because Cargo has no optional dev-dependencies; test gated with
  `#![cfg(feature = "e2e-web")]` so plain builds/tests are unaffected).
* In-process daemon on an OS-assigned port (concurrent-test safe) + headless
  Chromium via playwright-rs with a deterministic context (1280×800, dark
  theme, `prefers-reduced-motion: reduce`, en-US/UTC). Tests:
  `chat_round_trip_with_tool_call` (session creation via the sidebar picker,
  streamed markdown response, `set_title` tool call reflected in the sidebar)
  and `reload_replays_history`.
* Screenshot goldens in `tests/web_snapshots/` via playwright's screenshot
  assertions: animations disabled, auto-retry until settled, **strict
  zero-pixel tolerance** (renders verified byte-stable across runs thanks to
  reduced motion + deterministic ids); `UPDATE_SNAPSHOTS=1` regenerates,
  mismatches write `-actual`/`-diff` PNGs. **The two golden PNGs need human
  visual review.**

## Web UI: reduced-motion support

* `theme.css`: global `@media (prefers-reduced-motion: reduce)` rule freezing
  CSS animations/transitions (accessibility win + deterministic screenshots).
* `Spinner.tsx`: honors the media query — renders one static frame instead of a
  `requestAnimationFrame` loop.

## check.bash

New guarded "Web UI e2e (Playwright)" step: runs only when npm and the
Playwright 1.60 chromium build are installed (skip note otherwise) and falls
back to a temp npm cache when `~/.npm` isn't writable. For old-glibc hosts
where the driver's bundled node can't run, a comment documents the supported
env-var override (`PLAYWRIGHT_SKIP_DRIVER_DOWNLOAD=1` +
`PLAYWRIGHT_NODE_EXE` + `PLAYWRIGHT_CLI_JS`, forwarded by cargo to both the
build script and the test binary) — verified working end-to-end on this host.


crates/infinity-agent-cli/tests/snapshots/e2e_daemon_tui__e2e_after_busy_finish.snap
```diff
@@ -0,0 +1,27 @@
+---
```
luckyworkama, line 1: maybe somewhere we would want to explicitly set the screen width and height for the tests so that these snapshots are stable in differing environments (maybe the testing thing already does that internally?)

luckyworkama (approved):

crates/infinity-agent-cli/tests/snapshots/e2e_daemon_tui__e2e_after_busy_finish.snap
shadaj: The test harness takes a `async fn spawn(cols: u16, rows: u16) -> (Self, MockModelController) {` to specify the terminal size!

PR #61, opened 2026-07-08
feat: support multimodal (image) tool results end-to-end

shadaj:
Models can now declare image input support, RAP tools can return images in
tool results, and the sandbox `read_file` returns image files as attached
image content.

## Provider protocol (`infinity-agent-core`)

* `ModelEntry` gains `supports_image_input: bool` (`#[serde(default)]`, so
  the remote provider socket protocol stays backward compatible).
* `ModelProvider` gains `supports_image_input(model_id) -> bool` with a
  default implementation that consults `list_models`.
* `HistoryManager::get_history` now takes a `supports_image_input: bool`
  parameter: when `false`, image tool-result content is replaced in place
  (during the existing message translation, no extra allocation pass) with
  the `IMAGE_OMITTED_PLACEHOLDER` text. `run_completion` resolves the
  capability once per completion and passes it through, so switching to an
  image-capable model retroactively exposes images kept in history.
* Bedrock provider: all Claude models declare `supports_image_input: true`.

## RAP protocol (`rap-protocol`)

* New `RapToolResultContent` enum (`text` / `image { data, mediaType }`,
  base64) and an optional `content: Vec<RapToolResultContent>` field on
  `RapToolResult`. `text` remains the required plain-text fallback.
* Documented the new field in `docs/docs/rap/spec/basic/tool-result.md`
  (and the capability flag in `docs/docs/infinity-runtime/model-providers.md`).

## Daemon (`infinity-daemon`)

* `convert_callback` builds the rig tool result from structured `content`
  when present (images become `ToolResultContent::Image` with base64 data
  and MIME-derived media type), falling back to `text` otherwise.

## Sandbox (`sandbox-core` / `sandbox-local`)

* `read_file` detects image files by extension (png/jpg/jpeg/gif/webp, no
  size limit), returning a describing text plus base64 image content; text
  files behave exactly as before. `with_sandbox` was made generic over the
  action payload to carry the extra content channel.
* Tool manifest description updated to mention image reads.

## Tests

* agent-core: mock-model tests that image tool results are passed through
  to image-capable models and replaced with the placeholder otherwise;
  remote provider round-trips `supports_image_input` over the socket.
* daemon: unit tests for RAP→rig content conversion (fallbacks, media types).
* sandbox-local: integration tests for PNG round-trip content, extension →
  media type mapping, and text reads returning no structured content.


shadaj:
Confirmed that this works with a local build!

crates/infinity-agent-cli/src/acp_server.rs
```diff
@@ -764,17 +764,20 @@ fn daemon_msg_to_session_update(
...
+                        Some((Some(vec![diff]), None))
                     }
                     rap_protocol::DisplaySegment::Text(t) => {
-                        (None, Some(serde_json::Value::String(t)))
+                        Some((None, Some(serde_json::Value::String(t))))
                     }
+                    // Unsupported display types (e.g. images) — fall through
+                    // to the next segment.
+                    _ => None,
                 })
```
MingweiSamuel, line 781: Should we log this?

crates/infinity-agent-cli/src/terminal.rs
```diff
@@ -561,7 +568,8 @@ where
                                         viewport.print_line_above(Line::from(vec![]))?;
                                     }
                                 }
-                                None => {
+                                // No supported segment — just the checkmark.
+                                _ => {
```
MingweiSamuel: Should we log this? Is this case normally hit before this PR?

MingweiSamuel: perhaps it should say unable to display image, for images? Or could use the Kitty Graphics Protocol lol

crates/infinity-provider-protocol/src/lib.rs
```diff
@@ -83,6 +88,24 @@ pub trait ModelProvider: Send + Sync {
...
+        match self.list_models().await {
+            Ok(models) => models
+                .iter()
+                .any(|m| m.model_id == model_id && m.supports_image_input),
+            Err(e) => {
+                tracing::warn!("failed to list models to check image support: {e}");
+                false
+            }
+        }
+    }
```
MingweiSamuel: Is there a reason we don't just use `ModelEntry` `m.supports_image_input` directly?

crates/sandbox-core/src/server.rs
```diff
@@ -16,11 +17,23 @@ use tracing;
...
+/// as a third element, for operations that can return multimodal results.
+type ContentResult = Result<
+    (
+        String,
+        Option<Vec<DisplaySegment>>,
+        Option<Vec<RapToolResultContent>>,
+    ),
+    SandboxError,
+>;
+
```
MingweiSamuel, line 53: Perhaps these deserve to be actual structs or an enum or something

crates/sandbox-core/src/server.rs
```diff
@@ -1247,19 +1273,67 @@ async fn handle_execute_command_streaming_inner<
...
+fn image_media_type(path: &str) -> Option<&'static str> {
+    let ext = Path::new(path).extension()?.to_str()?.to_lowercase();
+    match ext.as_str() {
+        "png" => Some("image/png"),
+        "jpg" | "jpeg" => Some("image/jpeg"),
+        "gif" => Some("image/gif"),
+        "webp" => Some("image/webp"),
+        _ => None,
+    }
+}
```
MingweiSamuel: Is this fragile if the file has the wrong or missing extension?

crates/sandbox-core/src/server.rs
```diff
@@ -1247,19 +1273,67 @@ async fn handle_execute_command_streaming_inner<
...
+                // text summary fallback for those that can't.
+                let display = vec![
+                    DisplaySegment::Image(rap_protocol::ImageContent {
+                        data: encoded.clone(),
+                        media_type: media_type.to_owned(),
+                    }),
+                    DisplaySegment::Text(format!(
+                        "Read image ({media_type}, {} bytes);",
+                        bytes.len()
+                    )),
```
MingweiSamuel, line 1354: Does this later get selected? Maybe the DisplaySegment::Image should be kept by the client and it should contain alt text directly

infinity-ui/src/components/MessageItem.tsx
```diff
@@ -67,10 +67,23 @@ export const MessageItem = memo(function MessageItem({
     case "tool_result": {
       // Find the first supported segment type to render
       const seg = item.segments.find(
-        (s): s is DisplaySegment => s.type === "diff" || s.type === "text",
+        (s): s is DisplaySegment =>
+          s.type === "diff" || s.type === "text" || s.type === "image",
       );
```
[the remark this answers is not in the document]
MingweiSamuel: seems like a good idea

crates/rap-protocol/src/lib.rs
```diff
@@ -53,13 +68,45 @@ pub fn build_display_segments(
...
+    Text { text: String },
+    /// An image, base64-encoded.
+    #[serde(rename = "image")]
+    Image {
+        /// Base64-encoded image bytes.
+        data: String,
+        /// MIME type of the image, e.g. `image/png`.
+        #[serde(rename = "mediaType")]
+        media_type: String,
+    },
```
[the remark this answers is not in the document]
MingweiSamuel: wrong

crates/rap-protocol/src/lib.rs
```diff
@@ -53,13 +68,45 @@ pub fn build_display_segments(
...
+///
+/// A [`RapToolResult`] carries either a plain `text` result or a structured
+/// `content` list (which may include images) for runtimes and models that
+/// support multimodal inputs.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(tag = "type")]
+pub enum RapToolResultContent {
+    /// Plain text content.
+    #[serde(rename = "text")]
+    Text { text: String },
```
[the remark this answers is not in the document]
MingweiSamuel: wrong

infinity-ui/src/components/MessageItem.tsx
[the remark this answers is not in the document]
shadaj: Nah, we can support them all.

crates/infinity-agent-cli/src/acp_server.rs
shadaj: No, it's not an "issue", more that we are just picking whatever input type the client prefers.

crates/infinity-agent-cli/src/terminal.rs
shadaj: Good call — the terminal now renders `✓ [image — not displayable in terminal]` when a tool result has an image segment but no text/diff segment (previously a bare checkmark). In practice `read_file` also emits a trailing text-summary segment, which the terminal prefers over the image, so this branch is only hit for image-only results from other RAP servers.

crates/infinity-provider-protocol/src/lib.rs
shadaj: Good point — fixed. Removed the `ModelProvider::supports_image_input` trait method (which re-listed models on every completion) and now thread the flag from the already-resolved `ModelEntry`: `process_batch`/`run_completion` take a `supports_image_input: bool`, the daemon passes `catalog.find(&round_model).supports_image_input` (re-resolved per round so it follows mid-session model switches), and the Lambda resolves it once from `list_models()`. No more redundant per-turn lookup.

crates/sandbox-core/src/server.rs
shadaj: Done — replaced the `ContentResult` tuple with a `ToolOutput { text, display_as, content }` struct that `invoke_handler` destructures, so the three fields are named.

crates/sandbox-core/src/server.rs
shadaj: Good catch — it was fragile. Switched to content-based detection: `read_file` now sniffs the leading magic bytes (`detect_image_media_type`) instead of the file extension, so a mislabeled (`logo.png` that's actually text) or extension-less image is classified correctly, and the MIME type comes from the actual content. Added a test (`read_file_detects_image_by_content_not_extension`) covering a PNG named `.txt`.

crates/sandbox-core/src/server.rs
shadaj: The trailing text segment does get selected: the terminal/ACP clients skip the `Image` segment (they can't render it) and fall back to this text summary, so it's the human-facing description for non-image-capable UIs (the web UI renders the `Image` segment). Putting alt text directly on `DisplaySegment::Image` is a nice alternative, but it'd mean every client has to know to surface it; keeping an explicit text fallback segment means unchanged clients already do the right thing. Happy to switch to embedded alt text if you'd prefer it as the canonical shape.

MingweiSamuel (approved):

PR #63, opened 2026-07-09
refactor(infinity-agent-core): don't commit turn data to history until the turn is completed.

luckyworkama:


shadaj (approved):
This is super good and way better than the previous approach. We can also use this mechanism in a future PR to handle situations where a tool result is too large. I think basically if the model errors out saying the result was too large, then we drop that tool result?

The one tricky part of this PR (related to the above) is that we _do_ flush the turn output when the model errors out with an unknown error. I think that is probably the more conservative approach for now, but I can imagine situations where that leads to wedging, so we probably want to think more carefully in a followup about when we should flush vs drop.

crates/infinity-agent-core/src/event_processor.rs
```diff
@@ -1145,42 +1174,52 @@ where
...
                             Err(Into::<BoxError>::into("Stream timed out"))
                         }
                     },
                 }?;
 
                 if cancelled {
                     tracing::info!("Completion cancelled");
-                    history.remove_trailing_reasoning();
+                    // Terminal: keep whatever streamed so the user sees it.
+                    history.flush_turn();
```
shadaj: Hmm, we might need to remove trailing reasoning here, actually. I'm not sure if model providers tolerate user input immediately following a reasoning block? This code executes when a user interrupts a model while it is processing.

We should either test that the model providers tolerate this, or add back the trimming.

crates/infinity-agent-core/src/event_processor.rs
```diff
@@ -1145,42 +1174,52 @@ where
...
+                        history.discard_turn();
                         yield CompletionEvent::Info("Stream error (unexpected end), retrying...".to_owned());
-                        tracing::warn!("Stream ended unexpectedly, removing trailing reasoning and retrying...");
+                        tracing::warn!("Stream ended unexpectedly, discarding partial turn and retrying...");
                         tokio::time::sleep(Duration::from_secs(1)).await;
                         retry_count += 1;
                         continue 'outer;
                     } else {
+                        // Giving up: preserve whatever streamed, then error.
+                        history.flush_turn();
```
shadaj: Same as a above, may want to trim reasoning.

crates/infinity-agent-core/src/event_processor.rs
```diff
@@ -1192,19 +1231,22 @@ where
...
+                            history.discard_turn();
                             yield CompletionEvent::Info("Stream error (unexpected end), retrying...".to_owned());
-                            tracing::warn!("Stream error (unexpected end), retrying...");
+                            tracing::warn!("Stream error (unexpected end), discarding partial turn and retrying...");
                             tokio::time::sleep(Duration::from_secs(1)).await;
                             retry_count += 1;
                             continue 'outer;
                         } else {
+                            // Giving up: preserve whatever streamed, then error.
+                            history.flush_turn();
```
shadaj: Same as above.

crates/infinity-agent-core/src/event_processor.rs
```diff
@@ -1145,42 +1174,52 @@ where
...
                             if is_thinking {
                                 is_thinking = false;
                                 yield CompletionEvent::ThinkingEnd;
                             }
                             retry_count += 1;
                             continue 'outer;
                         } else {
+                            // Giving up: preserve whatever streamed so the user
+                            // still sees it, then surface the error.
+                            history.flush_turn();
```
shadaj: Same as below.

shadaj (changes requested):
I realized there is one concern wrt trailing reasoning.

shadaj (approved):
Awesome!

PR #72, opened 2026-07-14
feat(infinity-agent-cli): add `infinity daemon restart` and post-update daemon check

shadaj:
## `infinity daemon restart`

* New `DaemonCommands::Restart` subcommand backed by `daemon_client::restart_daemon()`:
  * Sends SIGTERM to the running daemon (found via the pid file, liveness-probed with signal 0 so stale pid files are ignored)
  * Waits up to 30s for the old process to fully exit — so its shutdown cleanup (removing the socket/pid files) can't race the new instance
  * Launches a fresh daemon via the existing `launch_daemon()` readiness handshake
  * If no daemon is running, just starts one
* `daemon stop` now uses the shared `running_daemon_pid()` helper instead of hand-parsing the pid file

## Post-update daemon check

* New `install::ensure_fresh_daemon()` runs after `infinity update`, `infinity rap update`, and `infinity provider update` (all of these update binaries the daemon executes or spawns):
  * If the daemon is **not** running → boots it, so the freshly installed version is live
  * If the daemon **is** running → prints a yellow warning that it's still on the previous version and to run `infinity daemon restart`
* `launch_daemon()` now resolves the executable via `daemon_exe_path()`, which strips the Linux `/proc/self/exe` " (deleted)" suffix — after a self-update, the old (replaced) binary would otherwise try to re-exec its deleted inode path; this ensures the freshly installed binary is spawned instead. Made `launch_daemon` `pub` for reuse from `install.rs`.

Verified with `./check.bash` (all checks pass, including web e2e).

### DIFF
--- a/crates/infinity-agent-cli/src/main.rs
+++ b/crates/infinity-agent-cli/src/main.rs
@@ -69,6 +69,8 @@ enum Commands {
 enum DaemonCommands {
     /// Stop the running daemon
     Stop,
+    /// Restart the daemon (stop it if running, then start a fresh instance)
+    Restart,
 }
 
 #[derive(clap::Subcommand, Debug)]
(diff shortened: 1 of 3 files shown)


crates/infinity-agent-cli/src/main.rs
```diff
@@ -178,18 +180,13 @@ async fn async_main(cli: Cli) -> Result<(), BoxError> {
...
-                    nix::sys::signal::kill(
-                        nix::unistd::Pid::from_raw(pid),
-                        nix::sys::signal::Signal::SIGTERM,
-                    )
-                    .map_err(|e| format!("failed to send SIGTERM: {e}"))?;
-                    println!("sent SIGTERM to daemon (pid {pid})");
+                    let pid = daemon_client::stop_daemon()
+                        .await?
+                        .ok_or("daemon is not running")?;
+                    println!("daemon stopped (pid {pid})");
```
MingweiSamuel, line 186: is this going to say "daemon stopped (pid daemon is not running)"?

MingweiSamuel: Ok I guess not since the `?`

MingweiSamuel (approved):

PR #75, opened 2026-07-14
fix(sandbox): detect repository root when the requested path is nested inside a repo (#65)

### LINKED ISSUE #65: Opening sandbox should identify repository root and include in tool response if different than requested path

shadaj:
* Add `sandbox_core::find_repo_root`, which walks up from a path to the
  closest ancestor containing a `.jj` directory or a `.git` entry
  (directory, or file for worktrees/submodules). This makes jj repos
  detectable when the cwd is inside a nested folder — jj subdirectories
  carry no marker of their own, so the previous `path.join(".jj")` check
  silently fell through to git mode (or failed entirely for
  non-colocated jj repos).
* `LocalBackend::init_repo` now resolves the requested path to the
  repository root before storing it as the remote URI, so sandboxes are
  always created from the repo root. Falls back to the path as given
  when no root is found, preserving existing error paths.
* `clone_repo` and `open_sandbox_direct` responses now include a note
  when the detected repository root differs from the requested path,
  telling the agent that the sandbox operates on the whole repository
  and that file paths are relative to the root.
* The migration-import cwd check also walks up to the repo root instead
  of requiring cwd to be the root itself.
* Add integration tests (`nested_repo_root.rs`) covering jj and git
  clones from nested directories (mode detection + root note + reads
  relative to the root) and the no-note case when cloning from the root.
  
Fixes #65

### DIFF
--- a/crates/sandbox-core/src/server.rs
+++ b/crates/sandbox-core/src/server.rs
@@ -469,8 +469,12 @@ async fn handle_clone_repo<B: SandboxBackend, M: MetadataStore, C: CallbackClien
     let bookmark = format!("sandbox-{}", invocation.group_id);
     let path = PathBuf::from(&remote_uri);
 
+    // If the requested path was nested inside the repository (rather than
+    // being the repo root itself), note the detected root in the response.
+    let root_note = repo_root_note(&args.repo, &path);
+
     // Resolve the base revision to an absolute jj change_id.
-    let (repo_state, msg) = if path.join(".jj").is_dir() {
+    let (repo_state, mut msg) = if path.join(".jj").is_dir() {
         // Jujutsu repo — resolve base revision via jj.
         let _ = run_jj(&path, &["workspace", "update-stale"]).await;
         let rev_to_resolve = args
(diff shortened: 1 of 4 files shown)


crates/sandbox-core/src/server.rs
```diff
@@ -558,7 +562,28 @@ async fn handle_clone_repo<B: SandboxBackend, M: MetadataStore, C: CallbackClien
     state.metadata.put(&repo_state).await?;
 
     tracing::info!(group_id = %invocation.group_id, remote = %remote_uri, "repo cloned");
-    Ok(msg)
+    Ok(match root_note {
+        Some(note) => format!("{msg}{note}"),
+        None => msg,
+    })
+}
```
MingweiSamuel: nit: prefer `if let Some(x) =` on options

crates/sandbox-core/src/server.rs
```diff
@@ -593,7 +618,11 @@ async fn handle_open_sandbox_direct<B: SandboxBackend, M: MetadataStore, C: Call
     state.metadata.put(&repo_state).await?;
 
     tracing::info!(group_id = %invocation.group_id, remote = %remote_uri, "opened direct sandbox");
-    Ok("Repository initialized (Direct mode — file edits require approval, commands run without file write access unless write-orig is granted).".to_owned())
+    let mut msg = "Repository initialized (Direct mode — file edits require approval, commands run without file write access unless write-orig is granted).".to_owned();
+    if let Some(note) = repo_root_note(&args.repo, Path::new(&remote_uri)) {
```
MingweiSamuel, line 627: what is the difference between repo and remote_uri here? Is it possible to have a direct sandbox on a non-VCS directory?

also nit: would be nice if the `msg ++ note` concatenation was the same in both methods

MingweiSamuel (approved):

crates/sandbox-core/src/server.rs
shadaj: Done — switched to `if let Some(note)` + `push_str`, matching the pattern in `handle_open_sandbox_direct` (979a80d).

crates/sandbox-core/src/server.rs
shadaj: `args.repo` is the path exactly as the caller requested it, while `remote_uri` is the backend-resolved repository root: `LocalBackend::init_repo` canonicalizes the path and walks up to the nearest `.jj`/`.git`. They differ precisely when the requested path was nested inside the repo, which is what `repo_root_note` detects.

And yes, a Direct sandbox on a non-VCS directory is possible (and was before this PR too): when no root is found, `init_repo` falls back to the requested path itself, so Direct mode still opens and `repo_root_note` returns `None` (its `canonicalize`/`starts_with` checks fail gracefully for git remote URIs as well). I added a comment in 979a80d spelling this out, and unified the `msg` + note concatenation between the two handlers.

PR #78, opened 2026-07-16
feat(sandbox-core,sandbox-local): add interfaces to extend with custom VCS backends

shadaj:


MingweiSamuel (approved):
How will this integrate with checkpointing? Similar mechanism?

shadaj:
@MingweiSamuel re-requesting review because I refactored so that Jj and Git are just also pluggable backends

MingweiSamuel (approved):
skimmed the diff

PR #82, opened 2026-07-21
fix(infinity-agent-core): suppress stream content after the turn's tool call

shadaj:
Confirmed bug: with Bedrock (Claude with adaptive/interleaved thinking), the model can
stream *concurrent* tool calls in a single assistant message, with reasoning
interleaved between them. `run_completion` only executes the first tool call
(the rest are ignored), but it kept forwarding everything streamed after it:

* the ignored calls' name/argument deltas and the interleaved reasoning were
  yielded as `ThinkingChunk`s, so clients (web UI, TUI, ACP) saw "thinking"
  streaming *after* the `ToolCall` but *before* its `ToolResult` display event
  (which only arrives in a later round, via the MCP/RAP callback)
* worse, a trailing reasoning/text block was committed to history *after* the
  tool call, which broke the `history.last()` match in `handle_content` when
  the tool result arrived — the result was silently dropped and the call
  stranded as unanswered

Fix in `run_completion`: once the turn's first tool call has been emitted,
drop all subsequent stream content for that turn (logging it at info level)
instead of buffering/yielding it; only `Final` still flows through to flush
the turn and finish the round. This keeps the executed tool call as the last
committed history entry and restores the expected client event order:
`ToolCall → ResponseDone → ToolResult → StartOutput → thinking…`.

Also removed the now-dead "Ignoring batched tool call" branch and normalized
the surrounding indentation (rustfmt does not format `async_stream` bodies).

Added regression test `concurrent_tool_calls_suppress_trailing_stream_content`
that replays the Bedrock chunk sequence (reasoning → tc-1 → tc-2 deltas →
interleaved reasoning → tc-2) and asserts no display events are emitted after
the executed call and that history still ends with tc-1.

### DIFF
--- a/crates/infinity-agent-core/src/event_processor.rs
+++ b/crates/infinity-agent-core/src/event_processor.rs
@@ -1291,123 +1291,134 @@ where
                 let completion_id = format!("{}-{}-completion-{}", group_id, message_id, completion_counter);
                 completion_counter += 1;
 
-                  if let StreamedAssistantContent::ToolCall { .. } = chunk && has_emitted_tool_call {
+                // After the turn's first tool call, the model may keep
+                // streaming: Bedrock emits *concurrent* tool calls in one
+                // assistant message (each closed content block yields
+                // another `ToolCall`), with interleaved reasoning between
+                // them. Only the first call is executed — its result
+                // arrives in a later round — so everything streamed after
+                // it must be dropped here. Forwarding it would surface the
+                // ignored calls' argument deltas and reasoning as
+                // "thinking" that reaches clients *before* the executed
+                // call's result, and committing trailing reasoning/text to
+                // history after the tool call would break the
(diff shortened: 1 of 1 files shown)


crates/infinity-agent-core/src/event_processor.rs
MingweiSamuel, line 1460: So the case is there may be multiple Final blocks when there are multiple tool calls?

MingweiSamuel (approved):

crates/infinity-agent-core/src/event_processor.rs
shadaj: Ah, no, this is mostly a UI issue that we would pass through tool call deltas _after_ the first tool call and so you would get a jumbled mess of a tool call invocation but then thinking in the web UI. So the simpler thing is just to skip all content after the first tool call until the `Final`.

PR #83, opened 2026-07-24
chore: ensure `check.bash` passes

MingweiSamuel:
On AL2023 dev machine.

updating playwright-rs to 0.14.1 (from 0.14.0) fixes the setup issue

test(infinity-daemon): tolerate minor rasterization drift in web e2e screenshots

The Playwright screenshot assertions used `max_diff_pixels(0)`, which failed on
hosts where font rasterization drifts by a few pixels at glyph edges (16–108
pixels after Playwright's default 0.2 per-pixel color threshold; at most 0.036%
of raw pixels, confirmed to be antialiasing clusters around text).

* Replace `max_diff_pixels(0)` with `max_diff_pixel_ratio(0.0005)` (0.05% of the
  frame) in `assert_screenshot` — ~4x headroom over the observed drift while
  still catching any real UI regression, which moves orders of magnitude more
  pixels. Ratio-based so it scales with snapshot size.
* Update the comment to document the observed drift and rationale.
* Remove stale `*-actual.png` / `*-diff.png` artifacts left behind by the
  failed runs in `tests/web_snapshots/`.

Verified: `cargo test -p infinity-daemon --features e2e-web --test web_e2e`
passes (5/5), and the full `./check.bash` passes.


shadaj (approved):
Nice very helpful!

PR #88, opened 2026-07-29
fix(infinity-agent-core): return tool failures to the agent

shadaj:
* Add a shared `send_tool_error` helper that enqueues an error `ToolResult` with the original tool and call IDs
* Convert thread-tool argument and relationship validation failures, including missing `close_thread.thread_id`, from propagated errors or duplicated message construction into queued tool results
* Make the batch processor enqueue a generic `Error: Tool call failed` result when asynchronous tool execution returns an error, while retaining the detailed display event and logging delivery failures
* Add regressions covering missing `close_thread` arguments and generic failed-tool fallback delivery


MingweiSamuel (approved):
Nice, didn't realize this wasn't already the case, will help a lot

PR #89, opened 2026-07-29
fix(infinity-agent-cli): don't eat scrollback when a resize races the re-anchor cursor query

shadaj:
When the terminal is resized continuously (e.g. dragging a pane divider in
Zed), the terminal keeps reflowing while the viewport's re-anchor cursor
query (`CSI 6n`) is in flight. Acting on the stale reply re-saved the anchor
too high (a growing reflow pulls scrollback rows down, moving the true anchor
with them) and the subsequent clear-from-anchor-down erased the tail of the
assistant output.

* `InlineViewport::re_anchor` now re-checks the terminal size after every
  cursor-query round-trip and, if the geometry moved again, absorbs the new
  resize and re-queries until stable
* `InlineViewport::draw` runs `re_anchor` before computing any
  geometry-dependent values (frame area, ideal viewport position), so a draw
  that absorbs a raced resize is fully consistent with the newest geometry
* Regression test `tui_resize_race.rs` reproduces the race with a `TermOut`
  interposer that resizes the emulator right after answering the cursor
  query (before the fix, 6 of 30 output lines were erased)
* Test harness: `TuiHarness::spawn_with_term` allows spawning the TUI
  against a caller-provided `TermOut` and event channel

### DIFF
--- a/crates/infinity-agent-cli/src/inline_viewport.rs
+++ b/crates/infinity-agent-cli/src/inline_viewport.rs
@@ -439,19 +439,38 @@ impl<T: TermOut> InlineViewport<T> {
         Ok(())
     }
 
-    /// Re-derive the anchor after a resize, with a single cursor query.
+    /// Re-derive the anchor after a resize, with a cursor query.
     ///
     /// See the type-level docs for the strategy. Ends with the live cursor
     /// parked on the (re-saved) anchor.
+    ///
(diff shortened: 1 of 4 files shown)


MingweiSamuel (approved):
Great!

MingweiSamuel:
Would be more legible if the in-flight method was properly async?

PR #90, opened 2026-07-30
fix(infinity-daemon): add `keeps_session_alive` flag

akainth015:
to prevent non-interactive clients from blocking idle shutdown

Add a `keeps_session_alive` boolean field to `ClientMessage::CreateSession` and `ClientMessage::Connect` (defaulting to `true` via serde for backward compat). Connections that set this to `false` are tracked but do not prevent the session from idling out — enabling persistent but passive client connections (e.g. bots) that should not keep sessions warm indefinitely.

Key changes:

* `infinity-protocol`: new `keeps_session_alive` field on `CreateSession` and `Connect`, with `#[serde(default = "default_true")]`.

* `client_handler.rs`: tracks `connection_keeps_alive` per connection and threads it through `attach_client` and `send_input`.

* `session/thread_worker.rs`: replaces bare `UnboundedSender` subscriber lists with a `Subscriber` struct carrying `tx` + `keeps_session_alive`. Idle-exit and has-clients checks now only consider keep-alive subscribers.

* `session/mod.rs`: exposes a `worker_shutdown` `CancellationToken` on sessions. `send_input` uses a retry loop (up to 2 attempts) to handle the race where the agent loop begins winding down between the restart check and the actual send — inputs arriving mid-wind-down now transparently restart the session.

* `infinity-agent-cli`: all existing call sites pass `keeps_session_alive: true` (preserving current behavior for interactive CLI clients).

* `tests/keep_alive.rs`: three integration tests covering non-keep-alive idle exit, keep-alive warmth, and input-after-idle-exit restart.

### DIFF
--- a/crates/infinity-daemon/src/session/mod.rs
+++ b/crates/infinity-daemon/src/session/mod.rs
@@ -34,7 +34,7 @@ pub use thread_worker::thread_worker;
 type BoxError = Box<dyn std::error::Error + Send + Sync>;
 
 /// Re-export from thread_worker.
-pub use thread_worker::{SubscribeRequest, ThreadSubscribers};
+pub use thread_worker::{SubscribeRequest, Subscriber, ThreadSubscribers};
 
 /// Message sent to the agent loop — either user input, a subscribe request,
 /// or a model switch for a specific thread.
@@ -274,8 +274,8 @@ impl SessionManager {
             let smap = session.subscriber_map.lock().expect("bug: mutex poisoned");
             if let Some(subs) = smap.get(group_id) {
                 let subs = subs.lock().expect("bug: mutex poisoned");
-                for tx in subs.iter() {
-                    let _ = tx.send(msg.clone());
+                for sub in subs.iter() {
+                    let _ = sub.tx.send(msg.clone());
                 }
             }
         }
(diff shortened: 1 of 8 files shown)


crates/infinity-daemon/src/session/thread_worker.rs
```diff
@@ -17,11 +17,20 @@ use crate::session::ActiveThreads;
...
+}
+
 /// Shared subscriber list for a thread worker.
-pub type ThreadSubscribers = Arc<std::sync::Mutex<Vec<mpsc::UnboundedSender<DaemonMessage>>>>;
+pub type ThreadSubscribers = Arc<std::sync::Mutex<Vec<Subscriber>>>;
 
-/// Subscribe request: (client_tx, want_replay).
-pub type SubscribeRequest = (mpsc::UnboundedSender<DaemonMessage>, bool);
+/// Subscribe request: (client_tx, want_replay, keeps_session_alive).
+pub type SubscribeRequest = (mpsc::UnboundedSender<DaemonMessage>, bool, bool);
```
shadaj: Should turn this into a struct with field name.

crates/infinity-daemon/src/session/mod.rs
```diff
@@ -525,11 +529,14 @@ impl SessionManager {
...
     /// clears the shut_down flag and starts a new agent loop first.
+    ///
+    /// `client_tx` pairs the client's sender with its `keeps_session_alive`
+    /// flag, used when re-attaching the client after a session restart.
     pub async fn send_input(
         &mut self,
         thread_id: &str,
         msg: (InputMessage, Option<String>),
-        client_tx: Option<mpsc::UnboundedSender<DaemonMessage>>,
+        client_tx: Option<(mpsc::UnboundedSender<DaemonMessage>, bool)>,
```
shadaj: Should use `Subscriber` here.

shadaj (approved):
Two fixup areas

PR #92, opened 2026-08-04
feat(infinity-agent-core)!: add high-level agent system API

shadaj:
## Stack

This is **PR 2 of 2**, based on the shared-engine refactor in #96. Review #96 first; this diff contains the application-facing API, local protocol adapters, and documentation.

## Summary

Add ergonomic local agent-system APIs on top of the engine extracted in #96:

- static builder conveniences for tools, prompts, and RAP notification;
- channel-backed `ThreadHandle`s for sending inputs and streaming events;
- launcher mode and `ThreadBuilder` for per-thread tools, prompts, and models;
- root-based configuration inheritance for child threads;
- direct local `McpToolSet` and `RapToolSet` adapters;
- usage-oriented high-level and low-level documentation.

The daemon, Lambda, stores, driver, thread pipeline, admission, lifecycle behavior, and embedding-oriented bridge foundations are reviewed in #96.

### `ThreadHandle` and handle mode

`system/local/handle.rs` adds a channel-based observer and registry:

- `RunningSystem::thread_handle(id)` attaches to a thread and returns a handle that can send inputs and receive its event stream.
- The registry is owned by the running system, so handles continue receiving events when a driver idles and respawns.
- Multiple handles can observe the same thread; dropping one prunes its registration without affecting the others.
- Subscription installation is acknowledged before attach completes, preserving attach-then-send ordering.

Tests cover sending and receiving, streaming across tool-call rounds, driver respawn, and pruning dropped handles.

### Launcher mode

`system/local/launch.rs` adds:

- `LocalAgentSystem::with_thread_launcher()` and `LaunchingSystem`;
- `ThreadBuilder` for launching a new root thread with its own tools, system prompt, and model;
- attachment to existing threads through launcher handles without implicitly creating them.

Launch configuration is registered before the seed message is sent. `UnionConfigSource` and `UnionModelSource` resolve entries by the thread's root ID, so child threads inherit their parent's launch-specific tools, prompt, and model while retaining system-wide configuration.

Tests cover tool/prompt union, child inheritance, per-thread model selection, and attach-only behavior.

### Builder typestate and conveniences

`LocalAgentSystem` gains explicit operation modes:

- `Handles` is the default and `start()` returns `RunningSystem` with built-in handles.
- `Launcher` is selected with `with_thread_launcher()` and `start()` returns `LaunchingSystem`.
- `start_with_observer` remains available for embeddings that provide their own observer.

`AgentSystemBuilder` also gains static `tool`, `tools`, `extra_system_prompt`, and `rap_notifier` helpers for applications that do not need a custom `ThreadConfigSource`.

### Local MCP and RAP adapters

- `McpToolSet` exposes a lazy stdio or Streamable HTTP MCP client as local list/invoke tools while sharing metadata and dispatch with the daemon-facing `McpClient`.
- `RapToolSet` discovers manifest tools and returns the shared core `RapTool` implementation. It takes an explicit callback URL rather than capturing the most recently invoked system sender.
- `RapCallbackBridge` is bound separately and explicitly attached with `serve_into(system.sender())`, making callback ownership and view-update handling unambiguous. There is no late-bound sender mutex or destination switching.

### Documentation

Adds **The Agent System API** section:

1. overview;
2. building a system;
3. running locally;
4. RAP servers;
5. MCP servers;
6. custom tools;
7. dynamic configuration;
8. observers;
9. step mode;
10. customizing the engine.

It also adds focused low-level guides for the history manager and completion loop, and updates existing runtime pages to direct application authors toward the new API.

## Validation

- `cargo fmt --all -- --check`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo test --workspace`, including doc tests

## Breaking changes

Relative to #96:

- `LocalAgentSystem` gains a mode type parameter, defaulting to `Handles`.
- The default local startup method is `start()`; custom observers continue to use `start_with_observer`.
- `ThreadHandle`, `HandleObserver`, `HandleSubscribeRequest`, `LaunchingSystem`, and `ThreadBuilder` are exported under `system::local`.
- `RapTool` gains an optional callback URL used by local RAP tool sets.

No daemon behavior changes are introduced in this PR.


MingweiSamuel:
Oh god

shadaj:
@MingweiSamuel I'm not sure how we want to go about reviewing this, most of the big blobs of code are unchanged from before, just moved around, but the new system builder, thread handle, etc APIs are new (and do have tests)

MingweiSamuel:
Whats the differences between this (higher level?) API and the APIs used by the daemon (cli?) directly?

shadaj:
> Whats the differences between this (higher level?) API and the APIs used by the daemon (cli?) directly?

The main difference is that the new high-level API allows for directly controlling individual agent instances using handles, whereas the daemon implementation uses its own `Observer` implementation so that it can stream events across all agents. So it primarily depends on whether the system you are building exposes the concurrent agents as is (use a custom `Observer`) or if you want to just launch a bunch of individual agents (this API).

docs/docs/infinity-runtime/low-level/overview.md
```diff
@@ -0,0 +1,67 @@
...
+pub trait InputSender: Send + Sync + Clone {
+    type Error: std::error::Error + Send + Sync + 'static;
+
+    /// Send a message to the input queue for processing.
+    async fn send_to_input_queue(
+        &self,
+        message: InputMessage,
+        group_id: &str,
+        dedup_id: &str,
+    ) -> Result<(), Self::Error>;
```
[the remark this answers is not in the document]
shadaj: Done. The trait snippet now takes only `message` and `dedup_id`, and the surrounding text explains that routing comes from `message.group_id`.

crates/infinity-mcp-bridge/Cargo.toml
```diff
@@ -5,13 +5,16 @@ edition = "2024"
 license = "Apache-2.0"
 
 [dependencies]
+infinity-agent-core = { path = "../infinity-agent-core" }
 rap-protocol = { path = "../rap-protocol" }
 
+rig-core = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 tokio = { workspace = true, features = ["io-util", "process"] }
```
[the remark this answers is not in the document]
shadaj: Done. The normal `tokio` dependency now explicitly enables `rt`, so the library's `tokio::spawn` call does not depend on dev-dependency or workspace feature unification.

docs/docs/infinity-runtime/agent-systems/custom-tools.md
```diff
@@ -0,0 +1,305 @@
...
+        metadata: None,
+        synthetic: None,
+        display_as: None,
+        subscription: false,
+    };
+
+    sender
+        .send_to_input_queue(message, &group_id, &id)
+        .await
+        .map_err(Into::into)
```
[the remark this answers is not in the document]
shadaj: Done. All custom-tool examples now pass only the message and deduplication ID; each `InputMessage` continues to carry its target in `group_id`.

docs/docs/infinity-runtime/deploying-on-lambda.mdx
```diff
@@ -17,7 +17,7 @@ The agent runs as a short-lived Lambda function. It starts when a message arrive
...
+Tools are independent HTTP services. The runtime POSTs an invocation to a tool's Function URL and moves on without waiting for a response. When the tool finishes, it POSTs the result to a RAP Receiver Lambda, which enqueues it back on the input queue. This closes the loop: the agent wakes up again, picks up the tool result, and continues the conversation. The receiver accepts `tool_result`, `oauth`, `subscription_event`, and `user_choice` callbacks; `view_update` [server callbacks](/docs/rap/spec/server/user-choice) are acknowledged and dropped, since views are a display feature of the embedded daemon runtime.
```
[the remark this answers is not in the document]
shadaj: Done. The `view_update` text now links directly to the view-updates server callback specification.

crates/infinity-agent-core/src/system/local/launch.rs
```diff
@@ -0,0 +1,562 @@
...
+        .map(|(id, _)| id.clone())
+        .unwrap_or_else(|| thread_id.to_owned()))
+}
+
+/// Wraps the system-wide [`ModelSource`]: threads launched with their own
+/// model (and the subagent threads they spawn, via root resolution) resolve
+/// through it; all other threads fall through to the inner source.
+pub(crate) struct UnionModelSource<C: ConversationStore, M: InputSender> {
+    pub(crate) inner: Box<dyn ModelSource>,
+    pub(crate) registry: Rc<LaunchRegistry<M>>,
```
MingweiSamuel: Seems awkward that LaunchRegistry is Rc wrapped, could it either be `Rc<RefCell<LaunchRegistry>>` or just have `threads: Rc<RefCell<HashMap<String, LaunchConfig>>>` inside the struct?

MingweiSamuel: Seems like every place it is used it is wrapped in `Rc`

MingweiSamuel: Putting the RC inside would match `HandleRegistry`

crates/infinity-agent-core/src/system/local/handle.rs
```diff
@@ -0,0 +1,392 @@
...
+use crate::system::events::{AgentEvent, ReplaySnapshot};
+use crate::system::observer::ThreadObserver;
+
+/// The handle subscribers attached to each thread. One registry is shared by
+/// every [`HandleObserver`] the system creates, so a subscription survives
+/// its thread's driver idling out and respawning.
+#[derive(Clone, Default)]
+struct HandleRegistry {
+    subscribers: Rc<RefCell<HashMap<String, Vec<mpsc::UnboundedSender<AgentEvent>>>>>,
+}
```
MingweiSamuel: The `HandleRegistry` struct doesn't seem to really do anything? Could just make `HandleObserver` have `subscriber_registry: Rc<RefCell<HashMap<String, Vec<mpsc::UnboundedSender<AgentEvent>>>>>` or however you want to name it. Will probably have to silence type complexity

crates/infinity-agent-core/src/system/builder.rs
```diff
@@ -102,12 +106,16 @@ where
     conversation_store: C,
     state_store: S,
     model: Box<dyn ModelSource>,
+    tools: Vec<Rc<dyn Tool<M>>>,
```
MingweiSamuel: Is the `Rc` needed? Could we just use `Box<dyn _>`?

crates/infinity-agent-core/src/system/local/launch.rs
```diff
@@ -0,0 +1,562 @@
...
+use crate::message::InputMessage;
+use crate::system::builder::{AgentSystem, Launcher, LocalAgentSystem};
+use crate::system::config::{ThreadConfig, ThreadConfigSource};
+use crate::system::model::{ModelSource, ResolvedModel};
+
+type BoxError = Box<dyn std::error::Error + Send + Sync>;
+
+/// The tools, prompt, and model a thread was launched with.
+struct LaunchConfig<M: InputSender> {
+    tools: Vec<Rc<dyn Tool<M>>>,
```
MingweiSamuel, line 30: Does `tools` really need `Rc`?

crates/infinity-agent-core/src/system/local/launch.rs
```diff
@@ -0,0 +1,562 @@
...
+/// and prompt are unioned onto the system-wide configuration for every
+/// completion the thread runs.
+pub struct ThreadBuilder<'a, C, S, H>
+where
+    C: ConversationStore,
+    S: StateStore,
+    H: HttpClient,
+{
+    system: &'a LaunchingSystem<C, S, H>,
+    tools: Vec<Rc<dyn Tool<ChannelSender>>>,
```
MingweiSamuel: same here

crates/infinity-agent-core/src/system/local/launch.rs
```diff
@@ -0,0 +1,562 @@
...
+where
+    C: ConversationStore + 'static,
+    M: InputSender + 'static,
+    H: HttpClient + 'static,
+{
+    async fn resolve(&self, thread_id: &str) -> Result<ThreadConfig<M, H>, BoxError> {
+        let mut config = self.inner.resolve(thread_id).await?;
+        let root_id = root_thread_id(&self.conversation_store, thread_id).await?;
+        if let Some(local) = self.registry.threads.borrow().get(&root_id) {
+            config.tools.extend(local.tools.iter().cloned());
```
MingweiSamuel, line 122: I guess this is (the only? place) where `Rc::clone` is needed for `tools`

crates/infinity-agent-core/src/system/builder.rs
```diff
@@ -366,20 +465,42 @@ where
...
+/// How the threads of a built local system are created and attached to.
+/// The default, [`Handles`], attaches to any thread ID on demand; switch to
+/// [`Launcher`] with
+/// [`with_thread_launcher`](LocalAgentSystem::with_thread_launcher) to create
+/// threads through a [`ThreadBuilder`](super::local::ThreadBuilder) with their own
+/// tools and prompt.
+pub struct Handles;
+
+/// Marker for launcher mode; see [`Handles`].
+pub struct Launcher;
```
MingweiSamuel: use empty enum if these are just marker types, to prevent accidental instantiation
```suggestion
pub enum Handles {}

/// Marker for launcher mode; see [`Handles`].
pub enum Launche {}
```

docs/docs/infinity-runtime/agent-systems/building-a-system.md
```diff
@@ -0,0 +1,78 @@
...
+
+# Build Your First Agent
+This quickstart creates a local agent with in-memory state, launches one conversation thread, and streams its events. You need a Tokio `LocalSet` because local agent tasks do not require `Send`.
+
+```rust
+use infinity_agent_core::stores::{InMemoryConversationStore, InMemoryStateStore};
+use infinity_agent_core::system::{AgentEvent, AgentSystemBuilder, StaticModel};
+use tokio::task::LocalSet;
+
+#[tokio::main]
```
MingweiSamuel, line 14: `#[tokio::main(flavor = "current_thread")]` can be used to automatically set up the local set, I believe

MingweiSamuel (approved):
How confident are you that this isn't over-engineered? It seems like there are a lot more code paths/features than are strictly necessary, at a glance

crates/infinity-agent-core/src/system/local/launch.rs
shadaj: Fixed. `LaunchRegistry` is now cloneable and owns an `Rc<RefCell<...>>` internally, matching the handle subscriber storage pattern. Callers no longer wrap every registry use in another `Rc`.

crates/infinity-agent-core/src/system/local/handle.rs
shadaj: Fixed. I removed `HandleRegistry`; `HandleObserver` now directly contains the shared `Rc<RefCell<HashMap<...>>>`, and the observer factory clones that shared map into each driver observer.

crates/infinity-agent-core/src/system/builder.rs
shadaj: Fixed. The system builder now stores `Vec<Box<dyn Tool<_>>>`; it converts each tool to `Rc` only when building `StaticThreadConfig`, where repeated per-thread resolution actually requires cloning shared tool instances.

crates/infinity-agent-core/src/system/local/launch.rs
shadaj: The stored launch configuration does need shared ownership: `UnionConfigSource::resolve` can run repeatedly for the root and its children, and every resulting `ThreadConfig` must own a reusable reference to the same tool instances. I kept `LaunchConfig::tools` as `Rc`, but moved the conversion later so `ThreadBuilder` itself stores `Box` until `launch()`.

crates/infinity-agent-core/src/system/local/launch.rs
shadaj: Fixed here as well. `ThreadBuilder::tools` is now `Vec<Box<dyn Tool<ChannelSender>>>`; `launch()` converts to `Rc` only when inserting the reusable launch configuration.

crates/infinity-agent-core/src/system/local/launch.rs
shadaj: Yes, this is the sharing boundary. A launch configuration is resolved repeatedly for the root and child threads, so each resolved `ThreadConfig` clones the stored `Rc` references. Builders now retain `Box` ownership and convert only when the configuration is registered.

crates/infinity-agent-core/src/system/builder.rs
shadaj: These marker types were removed entirely when the implicit-create API was dropped. `LocalAgentSystem` no longer has a mode generic, and `build_local().start()` directly returns the thread-builder-based system.

docs/docs/infinity-runtime/agent-systems/building-a-system.md
shadaj: I checked Tokio's macro expansion: `flavor = "current_thread"` builds a normal current-thread runtime and does not enter a `LocalSet`; only Tokio's separate local runtime does that. Since `LocalAgentSystem::start` uses `spawn_local`, removing `LocalSet::run_until` would panic, so I kept the explicit `LocalSet` example.

PR #94, opened 2026-08-06
fix(web): stop replays from duplicating the transcript on reconnect

shadaj:
* `infinity-web/src/App.tsx`: a `Replay` always carries the complete
  transcript for the connected thread, so the handler now clears the
  message state (bumping the generation to invalidate MessageList's
  prefix cache), pending choices, and streaming flag before processing
  the replayed history. Previously the WS-reconnect path (`Welcome` →
  re-`Connect`) never reset the transcript — unlike `navigateTo` /
  `MigrateComplete` — so each reconnect appended a full extra copy of
  the history to the messages array and DOM, making the page
  progressively laggier with every disconnect/reconnect cycle. The
  input draft is deliberately preserved (unlike `resetMessages`).

* `crates/infinity-daemon/src/session/thread_worker.rs`:
  `handle_subscribe` now replaces an existing subscription on the same
  channel (`same_channel`) instead of pushing a duplicate. The web
  client re-sends `Connect` every 5s until `Connected` arrives; when a
  resume was slow, each retry stacked another subscriber for the same
  client, causing every display event (and another full `Replay`) to be
  delivered N times from then on.

### DIFF
--- a/crates/infinity-daemon/src/session/thread_worker.rs
+++ b/crates/infinity-daemon/src/session/thread_worker.rs
@@ -413,13 +413,17 @@ pub async fn thread_worker(
                 });
             }
         }
-        subscribers
-            .lock()
-            .expect("bug: mutex poisoned")
-            .push(Subscriber {
-                tx,
-                keeps_session_alive,
-            });
+        let mut subs = subscribers.lock().expect("bug: mutex poisoned");
+        // A client can legitimately re-send Connect on the same connection
+        // (e.g. its connect-retry timer fires while a slow resume is still in
+        // flight). Replace any existing subscription on the same channel
+        // instead of stacking a duplicate, which would deliver every display
+        // event to that client N times from then on.
+        subs.retain(|sub| !sub.tx.same_channel(&tx));
+        subs.push(Subscriber {
+            tx,
+            keeps_session_alive,
+        });
     };
 
     loop {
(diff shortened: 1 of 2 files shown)


crates/infinity-daemon/src/session/thread_worker.rs
```diff
@@ -413,13 +413,17 @@ pub async fn thread_worker(
...
+        // A client can legitimately re-send Connect on the same connection
+        // (e.g. its connect-retry timer fires while a slow resume is still in
+        // flight). Replace any existing subscription on the same channel
+        // instead of stacking a duplicate, which would deliver every display
+        // event to that client N times from then on.
+        subs.retain(|sub| !sub.tx.same_channel(&tx));
+        subs.push(Subscriber {
+            tx,
+            keeps_session_alive,
+        });
```
MingweiSamuel, line 426: Would `sub.tx.same_channel` actually be true in the re-connect case?

MingweiSamuel (approved):

crates/infinity-daemon/src/session/thread_worker.rs
shadaj: Not if it's a reconnect on a new TCP channel, but yes if the client just sends a second `Connect` to the same thread ID.

PR #95, opened 2026-08-10
feat: Show a placeholder in the session picker when there are no sessions

MingweiSamuel:
Previously, opening the session picker (/load or Ctrl+L) with no sessions
rendered a blank area with no indication of what was happening.

- session_picker.rs: render a dim italic placeholder line ("No sessions to
  load — press esc to dismiss and start a new one.") when the session list
  is empty; reserve one visible row for it via visible_rows(); make Enter
  dismiss the empty picker instead of doing nothing.
- terminal.rs: guard against an out-of-bounds panic when a sessions-updated
  notification arrives while the picker is open with an empty list (was
  indexing sessions[selected] unconditionally).
- tests: new snapshot test empty_session_picker_shows_placeholder covering
  both the placeholder rendering and Enter-to-dismiss behavior.

### DIFF
--- a/crates/infinity-agent-cli/src/session_picker.rs
+++ b/crates/infinity-agent-cli/src/session_picker.rs
@@ -45,7 +45,8 @@ impl SessionPicker {
     }
 
     pub fn visible_rows(&self) -> u16 {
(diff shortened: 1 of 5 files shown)


shadaj (approved):
Nice

PR #96, opened 2026-08-13
refactor!: extract shared agent system engine

shadaj:
## Stack

This is **PR 1 of 2**. Review and merge this PR first. The additive high-level API is split into #92, which is based on this branch.

## Summary

Extract the daemon's agent execution machinery into a shared `infinity_agent_core::system` engine, migrate the daemon and Lambda runtimes onto it, and extract the protocol components those embeddings share.

This PR intentionally stops at the engine and embedding boundary. Static builder conveniences, local MCP/RAP tool-set adapters, `ThreadHandle`, launcher mode, per-thread launch configuration, and the new usage guides are introduced by #92.

## Review guide

Most of this diff is relocation. The old daemon `thread_worker.rs` and `agent_loop.rs` loops, plus the old core batch processor, are replaced by focused core modules. The migrated daemon tests in `system/tests.rs` and their unchanged snapshots are the fidelity evidence for the port.

### Shared engine

- `system/local/driver.rs` ports the worker loop: batching, cancellation, deferral, compaction, tool-result waiting, and idle exit.
- `system/thread.rs` owns the completion pipeline, commit ordering, observer calls, config resolution, and tool dispatch.
- `system/local/router.rs` owns input routing, driver spawning, event admission, subscriptions, lifecycle notifications, and shutdown.
- `system/builder.rs` exposes embedding-oriented configuration, local observer startup, and batch-shaped `AgentSystem::step` for serverless platforms.
- `stores.rs` centralizes ancestor, compaction, deduplication, snapshot, and restore behavior. The daemon store wraps it with persisted extras.

### Behavior worth reviewing closely

- **Split event admission:** event-style input cannot create or resume a thread. Before spawning a driver, the router asks `ConversationStore::thread_exists` whether the exact root or child target exists, then asks `StateStore::is_thread_stopped` whether the embedding has explicitly stopped it. Unknown and stopped targets are dropped. User text bypasses both checks because it creates new threads and resumes stopped ones. Store failures fail open so a transient read cannot discard a real tool result.
- **Lifecycle-owned session state:** daemon input submission now always enqueues and contains no session-status policy. A `Live` lifecycle event resolves the thread's root session and clears both its idle and stopped flags. This applies equally to root and child activity and makes lifecycle the single source of runtime activity state. Event-style input for a stopped session never produces `Live` because router admission drops it first.
- **Lifecycle and subscription-aware activity:** `RunningSystem::thread_lifecycle` emits `Live` and `Idle` at the same points the router mutates its live set. The daemon keeps subscribed threads active for RAP server lifetime even though their drivers may idle.
- **Idle-exit races:** drivers and the router share a `LocalSet`; the final no-await channel checks make exit atomic with respect to routing. Input either reaches the existing driver or observes closed channels and causes a respawn.
- **Observer ordering:** history sync completes before `CompletionFinished` and tool dispatch. Failed dispatch enqueues an error tool result so the agent can recover.
- **Batch-shaped step mode:** one delivery is partitioned by `group_id`, deferral is applied sequentially per thread, and thread slices run concurrently.

### Protocol foundations

- `infinity-rap-bridge` provides callback conversion, prepared deduplication IDs, callback serving, and view-update forwarding for daemon and Lambda embeddings.
- `infinity-mcp-bridge` provides lazy stdio and Streamable HTTP clients plus canonical metadata and dispatch used by the daemon proxy.
- `invoke_rap_tool` is the shared RAP invocation path used by core, Lambda, and daemon tools, including a uniform non-2xx error result.
- Application-facing `RapToolSet` and `McpToolSet` adapters are deferred to #92.

### Production migrations

- **Daemon:** one daemon-lifetime agent system replaces per-session loops and restart-on-send behavior. RAP servers boot lazily from thread configuration and follow lifecycle/subscription activity. All input paths enqueue through the same router; lifecycle transitions own stopped/idle state updates.
- **Lambda:** a batch is processed through one `AgentSystem::step`; per-session tools resolve through `ThreadConfigSource`; the Rust `rap-receiver` uses shared callback conversion and replaces the JavaScript receiver. DSQL thread-hierarchy records provide exact event-target existence checks.
- **CLI:** `DisplayEvent` moves to the CLI because rendering is no longer a core engine concern.

## Diff orientation

The new core system modules include the migrated daemon fidelity tests and shared fixtures. The daemon itself shrinks substantially because `agent_loop.rs` and `thread_worker.rs` are removed. Application conveniences and the documentation rewrite are kept out of this PR so reviewers can focus on the engine and production migrations.

## Validation

- `cargo fmt --all -- --check`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo test --workspace`

## Breaking changes

- Removes `batch_processor` (`process_batch`, `process_input_item`, and core `DisplayEvent`) in favor of `AgentSystem::step` and observer-based local execution.
- `Thread` is internal and `AgentSystem` is not `Clone`.
- `ThreadObserver` replaces inline daemon persistence/display hooks; it has no `on_commit`.
- `EventCollector::take` returns `(thread_id, event)` pairs.
- `ConversationStore` gains `thread_exists`, which checks exact root or child records without creating them.
- `StateStore` gains the provided `is_thread_stopped` policy hook. User text may resume stopped threads; event-style input may not.
- Builder tools are stored as `Rc`; `Tool` gains defaulted `is_passive`.
- `ToolContext` and its builder lose `input_queue_arn`.
- Resident runtime types such as `RunningSystem`, `SubscribeHandle`, `ChannelSender`, and `ChannelSendError` live under `system::local`.
- The daemon no longer exports its old worker/loop/session implementation modules, sleep/RAP tool wrappers, or `boot_rap_servers`.
- `SessionManager::send_input` no longer accepts a thread ID or `user_driven` flag and performs no status admission; router admission and lifecycle events own those concerns.
- `rap_callback::start_callback_server` is replaced by `infinity_daemon::launch_session_manager`; callback serving accepts a `RapCallbackBridge`.
- `SessionManager::switch_model` accepts the requester's sender and returns `Result<(), String>`; `SharedSessionManager` is `Rc`.
- CLI `DisplayEvent` lives in `infinity_agent_cli::display`.


crates/infinity-agent-core/src/stores.rs
```diff
@@ -0,0 +1,553 @@
...
+            s.remove(tool_call_id);
+        }
+        Ok(())
+    }
+
+    async fn should_wake_thread_for_event(&self, thread_id: &str) -> Result<bool, Self::Error> {
+        let Some(known_threads) = &self.known_threads else {
+            // Unlinked stores cannot know which threads exist; admit the
+            // event rather than dropping input for a real thread.
+            return Ok(true);
```
MingweiSamuel: Seems like maybe we should eliminate the `None` case (see comment above in this file)

crates/infinity-agent-core/src/stores.rs
```diff
@@ -0,0 +1,553 @@
...
+pub struct InMemoryStateStore {
+    #[expect(clippy::type_complexity, reason = "shared state")]
+    processed_ids: Arc<Mutex<HashMap<String, (HashSet<String>, HashSet<String>)>>>,
+    metadata: Arc<Mutex<HashMap<String, serde_json::Value>>>,
+    subscriptions: Arc<Mutex<HashMap<String, HashSet<String>>>>,
+    /// Thread registry of the paired [`InMemoryConversationStore`], when
+    /// linked via [`for_conversations`](Self::for_conversations). Used by
+    /// [`should_wake_thread_for_event`](StateStore::should_wake_thread_for_event)
+    /// to distinguish real threads from stale event targets.
+    known_threads: Option<Arc<Mutex<HashMap<String, ThreadInfo>>>>,
```
MingweiSamuel: When is `Self::new()` used, resulting in no known threads? We should try to eliminate these weird edge cases if they're just for testing or whatnot

crates/infinity-daemon/src/memory_store.rs
```diff
@@ -21,17 +38,22 @@ impl std::fmt::Display for MemoryError {
...
 pub struct InMemoryConversationStore {
-    /// session_id -> ordered messages
-    #[expect(clippy::type_complexity, reason = "shared state")]
-    messages: Arc<Mutex<HashMap<String, Vec<(InfinityMessage, String)>>>>,
-    /// thread_id -> ThreadInfo
-    threads: Arc<Mutex<HashMap<String, ThreadInfo>>>,
-    /// thread_id -> compaction summaries
-    compaction_summaries: Arc<Mutex<HashMap<String, Vec<CompactionSummary>>>>,
+    /// The shared store semantics (threads, messages, compaction summaries).
+    core: core_stores::InMemoryConversationStore,
```
MingweiSamuel, line 51: Confusing to have two layers of `InMemoryConversationStore`

crates/infinity-daemon/src/memory_store.rs
```diff
@@ -1004,112 +938,73 @@ impl ConversationStore for InMemoryConversationStore {
...
-
 #[derive(Clone)]
 pub struct InMemoryStateStore {
-    #[expect(clippy::type_complexity, reason = "shared state")]
-    processed_ids: Arc<Mutex<HashMap<String, (HashSet<String>, HashSet<String>)>>>,
-    metadata: Arc<Mutex<HashMap<String, serde_json::Value>>>,
-    /// Per-thread active subscriptions: thread_id → set of tool_call_ids.
-    subscriptions: Arc<Mutex<HashMap<String, HashSet<String>>>>,
+    /// The shared store semantics (processed IDs, metadata, subscriptions).
+    core: core_stores::InMemoryStateStore,
```
MingweiSamuel, line 947: Likewise confusing to have nested types with the same name

crates/infinity-agent-core/src/system/local/driver.rs
```diff
@@ -0,0 +1,488 @@
...
+    },
+    /// The thread's driver exited: nothing is queued and no tool result is
+    /// on its way, so the thread is idle until the next message. The thread
+    /// may still hold active subscriptions; a subscription event respawns
+    /// its driver on arrival.
+    Idle {
+        /// The thread whose driver exited.
+        thread_id: String,
+    },
+}
```
MingweiSamuel: I would prefer to represent this as
```rust
pub struct ThreadLifecycleEvent { // or maybe ThreadLifecycleTransition`
    pub thread_id: String,
    pub lifecycle_state: ThreadLIfecycleState,
}

pub enum ThreadLIfecycleState {
    Live, Idle
}
```
assuming that structuring makes sense

crates/infinity-agent-core/src/system/local/router.rs
```diff
@@ -0,0 +1,547 @@
...
+    sender: ChannelSender,
+    subscribe_tx: mpsc::UnboundedSender<SubscribeMessage<Sub>>,
+    active_threads: ActiveThreads,
+    /// Receives a [`ThreadLifecycleEvent`] each time a thread's driver
+    /// spawns (`Live`) or exits (`Idle`: no pending tool call and no active
+    /// subscription). Embeddings use this to track per-conversation activity
+    /// and release resources — e.g. the Infinity Code daemon marks a session
+    /// active while any of its threads is live and shuts down the session's
+    /// RAP servers once none are and no keep-alive client is attached.
+    pub thread_lifecycle: mpsc::UnboundedReceiver<ThreadLifecycleEvent>,
```
MingweiSamuel: Suspicious to have just a single `pub` field, especially a channel receiver - can we encapsulate?

crates/infinity-agent-core/src/system/local/router.rs
```diff
@@ -0,0 +1,547 @@
...
+
+                running.send_user_text("t1", "hello").await;
+                let _req = ctrl.next_request().await;
+                ctrl.send_text("hi");
+                ctrl.finish();
+                collect_until_finished(&mut rx).await;
+                wait_idle(&mut running).await;
+                // Drain transitions from the first driver so the assertions below
+                // observe only the respawned one.
+                while running.thread_lifecycle.try_recv().is_ok() {}
```
MingweiSamuel: This is a bit suspicious and might indicate that a different synchronization structure would be better?

MingweiSamuel: I guess this is fine for a test

crates/infinity-agent-core/src/system/local/driver.rs
```diff
@@ -0,0 +1,488 @@
...
+            // be processed even if nothing else arrives.
+            let defer_ready = !defer.is_empty() && thread.pending_active_tool_call().is_none();
+
+            if batch.is_empty() && !defer_ready {
+                // From here to `return` there must be no await: the router
+                // runs on the same `LocalSet`, so a section without awaits is
+                // atomic with respect to it. Anything already in the channels
+                // is found by the `try_recv`s below, and anything sent after
+                // `return` finds the channels closed (dropping the receivers
+                // closes them), which makes the router respawn a driver.
```
MingweiSamuel: Should we put this in a non-async helper method (if there are not way too many contextual variables to pass in) to enforce no-await?

crates/infinity-agent-core/src/system/local/driver.rs
```diff
@@ -0,0 +1,488 @@
...
+                while let Ok((req, ack)) = subscribe_rx.try_recv() {
+                    observer.on_subscribe(&thread_id, req, thread.replay_snapshot());
+                    let _ = ack.send(());
+                }
+                if let Ok(item) = rx.try_recv() {
+                    batch.push(item);
```
MingweiSamuel: Comment: Though we could go to sleep, but because things have appeared in the buffer, we should not actually go to sleep. (Does this branch ever happen normally? I guess yes due to small races?)

crates/infinity-agent-core/src/system/local/driver.rs
```diff
@@ -0,0 +1,488 @@
...
+        if batch.is_empty() {
+            // Everything was deferred (or nothing arrived); wait for more input.
+            continue;
+        }
+
+        let (cancel_tx, cancel_rx) = oneshot::channel::<()>();
+        in_flight = Some(InFlightStep {
+            fut: Box::pin(thread.step_no_defer(batch, &observer, cancel_rx)),
+            cancel_tx,
+        });
```
MingweiSamuel, line 315: What if there is already `Some` in `in_flight`? Would have to be from the `req = subscribe_rx.recv(), if !subscribe_closed => {` case above

crates/infinity-agent-core/src/system/local/driver.rs
```diff
@@ -0,0 +1,488 @@
...
+
+type StepFuture<'a> = Pin<Box<dyn Future<Output = Result<StepOutcome, BoxError>> + 'a>>;
+
+/// An in-flight step paired with its cancellation handle. Keeping the two in
+/// one value makes it impossible for them to desynchronize: interrupting the
+/// step consumes both, and dropping the pair (e.g. on driver teardown) drops
+/// the sender, which the step observes as a cancellation.
+struct InFlightStep<'a> {
+    fut: StepFuture<'a>,
+    cancel_tx: oneshot::Sender<()>,
```
MingweiSamuel, line 40: Should this be a https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html

crates/infinity-agent-core/src/system/local/driver.rs
MingweiSamuel: I guess not due to the `continue`

crates/infinity-agent-core/src/system/local/driver.rs
```diff
@@ -0,0 +1,488 @@
...
+    let mut pending_batch: Vec<(InputMessage, String)> = Vec::new();
+    let mut compaction_triggered = false;
+    let mut total_tokens: u64 = 0;
+    // Set when subscribe_rx is closed, so selects stop polling it (a closed
+    // channel is always ready and would otherwise spin).
+    let mut subscribe_closed = false;
+
+    let mut in_flight: Option<InFlightStep<'_>> = None;
+
+    loop {
```
MingweiSamuel, line 141: We should clean up the control flow of this loop in a follow up

crates/infinity-agent-core/src/system/observer.rs
```diff
@@ -0,0 +1,107 @@
...
+///   exactly-once delivery for free: an event is either already reflected in
+///   the [`ReplaySnapshot`] a new subscriber receives, or it is broadcast to
+///   that subscriber afterwards — never both, never neither.
+/// - **Durability**: the async methods are awaited inline at their exact
+///   transition points, before the step continues past them, so state an
+///   implementation persists there is durable before the agent can react.
+///
+/// Implementations that need mutable state should use interior mutability
+/// (`RefCell`); the methods take `&self` so the runtime can hold shared
+/// references across a step.
```
MingweiSamuel, line 27: I suppose Infinity could wrap in RefCell itself, if this is always expected to be the case

crates/infinity-agent-core/src/event_processor.rs
```diff
@@ -1071,6 +1058,88 @@ where
...
+    },
+}
+
+/// Compute the [`InputEcho`] for an input that was just accepted into
+/// history. Returns `None` for inputs with no display representation (e.g. a
+/// synthetic event whose originating tool call is no longer in history).
+pub fn input_echo<C, S>(
+    current_history: &HistoryManager<C, S>,
+    input_msg: &InputMessage,
+) -> Option<InputEcho>
```
MingweiSamuel: Should this not just map to `AgentEvent` directly?

MingweiSamuel (approved):

crates/infinity-daemon/src/memory_store.rs
shadaj: Done. Renamed the daemon wrapper to `PersistentConversationStore` while retaining `InMemoryConversationStore` for the core in-memory implementation, so the two layers are now distinct at call sites.

crates/infinity-daemon/src/memory_store.rs
shadaj: Done. Renamed the daemon wrapper to `PersistentStateStore` while retaining `InMemoryStateStore` for the core in-memory implementation.

crates/infinity-agent-core/src/system/local/driver.rs
shadaj: Done. Lifecycle transitions are now represented as `ThreadLifecycleEvent { thread_id, state }`, with `ThreadLifecycleState::{Live, Idle}` carrying the state.

crates/infinity-agent-core/src/system/local/router.rs
shadaj: Done. The lifecycle receiver is now private and exposed through `next_lifecycle_event` and `try_next_lifecycle_event`; launcher mode delegates through its own lifecycle method.

crates/infinity-agent-core/src/system/local/driver.rs
shadaj: Done. Extracted the final idle check into synchronous `check_before_idle`, which makes the no-await requirement explicit for this atomic section.

crates/infinity-agent-core/src/system/local/driver.rs
shadaj: Done. Added a comment explaining that input can arrive between the initial drain and final synchronous check, so the existing driver continues without publishing a transient idle transition.

crates/infinity-agent-core/src/event_processor.rs
shadaj: Done. Removed `InputEcho`; `input_event` now returns `Option<AgentEvent>` directly, and the thread forwards the returned event.

MingweiSamuel (approved):

PR #105, opened 2026-08-26
perf(infinity-agent-core)!: cut idle-agent memory ~30% (154 → ~107 KB/agent)

shadaj:
Optimizes per-agent resident memory for the `agent_scale` benchmark added by the parent commit, in three independent pieces:

## Router owns driver futures directly (no `spawn_local` per driver)

* `route_loop` now drives all thread drivers through one `FuturesUnordered` pool instead of spawning each as its own `LocalSet` task. When a driver goes idle, its future yields its thread ID and the router immediately drops both the future's memory and the worker entry (input/subscribe channels).
* Previously a finished driver's `JoinHandle`, task allocation, and channel blocks were retained in the `workers` map until the thread's *next* message — ~13 KB per idle agent.
* Panic isolation is preserved (each pooled future is still wrapped in `rap_protocol::log_panic`); shutdown wind-down now just drains the pool.

## `InfinityMessage::SubscriptionEvent` payloads boxed

* Boxed `result` and `invocation` in the rare `SubscriptionEvent` variant, shrinking `size_of::<InfinityMessage>()` from 352 → 184 bytes. Every stored history message paid for the fattest variant inline. `Box` is serde-transparent, so the wire/persisted format is unchanged.
* Added `InfinityMessage::tool_result()` helper (the boxed field made cross-variant or-patterns impossible).
* (Boxing only `invocation` was measured and rejected: it *grows* the enum to 200 bytes because `SubscriptionEvent` with an inline result becomes the largest variant again.)

## Tool-call dedup derived from history instead of a durable index (BREAKING)

* `HistoryManager` no longer maintains `processed_tool_calls` / `pending_complete_tool_calls`. Incoming tool results are deduplicated by walking the history tail just in time: scan back across trailing tool calls/results (future-proof for concurrent calls, e.g. `tc tc tr tr`), accept on a matching unanswered call, reject as duplicate on a matching result, and discard as stale on any other message (user text, assistant content, subscription events — all turn boundaries, since a subscription event can only be injected once pending calls are settled).
* `safe_spawn_point` uses the same walk (tracking answered calls) instead of the set.
* Durable message-ID dedup is now limited to inputs that are not naturally idempotent: user text and subscription events (a redelivered subscription event would mint a fresh injected invocation). Tool results and assistant/tool-call items no longer persist IDs.
* **BREAKING**: `StateStore::get_processed_ids` now returns a single `HashSet<String>` and `add_processed_tool_calls` is removed — updated `InMemoryStateStore`, the daemon's `PersistentStateStore`, and the Lambda `DynamoDbStateStore` (old DynamoDB `processed_tool_calls` attributes are simply ignored). `ThreadState` keeps its `processed_tool_call_ids` field so old daemon snapshots still deserialize.

## Benchmark & docs

* `agent_scale` example now drains lifecycle notifications per wave (like a real embedding), so they aren't counted as per-agent memory.
* Landing page: refreshed `MemoryChart` data from a post-optimization 50k-agent run (6.08 GB vs 7.90 GB before; 118.8 KB/agent at that point, ~107 KB/agent after the dedup changes) and updated the copy; updated the history-manager doc for the new dedup model.
* Note: an extended 80k-agent benchmark run was in flight to measure the exact 8 GB crossing (~72k agents projected); `MemoryChart` data and copy may deserve a refresh from that run. Also left open: making the per-message dedup ID in `ConversationStore::append_messages` optional (user/subscription messages only), which needs a decision about the Lambda DSQL `message_id NOT NULL` column and daemon session-file compat.

All workspace tests pass (`cargo test --workspace`), clippy clean.

### DIFF
--- a/crates/infinity-agent-core/src/event_processor.rs
+++ b/crates/infinity-agent-core/src/event_processor.rs
@@ -128,6 +128,17 @@ pub struct PendingItem {
     message_id: String,
 }
 
+/// Result of matching an incoming tool result against the history tail (see
+/// [`HistoryManager::match_tool_result`]).
+enum ToolResultMatch {
+    /// A matching, still-unanswered tool call exists: the result is fresh.
+    Unanswered,
+    /// The matching call already has a result in history: duplicate delivery.
+    AlreadyAnswered,
+    /// No matching call in the live tail: the result is stale.
+    NoPendingCall,
+}
+
 pub struct HistoryManager<C: ConversationStore, S: StateStore> {
     conversation_store: C,
     state_store: S,
@@ -136,14 +147,12 @@ pub struct HistoryManager<C: ConversationStore, S: StateStore> {
     ancestor_chain: Vec<String>,
     pub history: RefCell<Vec<InfinityMessage>>,
     processed_message_ids: RefCell<HashSet<String>>,
-    processed_tool_calls: RefCell<HashSet<String>>,
     metadata: RefCell<Option<serde_json::Value>>,
     pending_items: RefCell<Vec<PendingItem>>,
     /// until a turn is complete the data lives here. If errors occur,
     /// it'll get discarded, if the turn completes, then it will get flushed to
     /// _both_ pending_items and history.
     turn_buffer: RefCell<Vec<PendingItem>>,
-    pending_complete_tool_calls: RefCell<HashSet<String>>,
     /// Tool call IDs that were interrupted by a new user message during
     /// `handle_content`. Callers can drain this via `take_interrupted_tool_calls`
     /// to send best-effort cancellation notifications to RAP tool servers.
@@ -187,10 +196,10 @@ impl<C: ConversationStore, S: StateStore> HistoryManager<C, S> {
             .await
             .unwrap_or(None);
 
-        let (processed_message_ids, processed_tool_calls) = state_store
+        let processed_message_ids = state_store
             .get_processed_ids(&thread_id)
             .await
-            .unwrap_or_else(|_| (HashSet::new(), HashSet::new()));
+            .unwrap_or_default();
 
         Ok(Self {
             conversation_store,
@@ -200,18 +209,16 @@ impl<C: ConversationStore, S: StateStore> HistoryManager<C, S> {
             ancestor_chain,
             history: RefCell::new(history),
             processed_message_ids: RefCell::new(processed_message_ids),
-            processed_tool_calls: RefCell::new(processed_tool_calls),
             metadata: RefCell::new(metadata),
             pending_items: RefCell::new(Vec::new()),
             turn_buffer: RefCell::new(Vec::new()),
-            pending_complete_tool_calls: RefCell::new(HashSet::new()),
             interrupted_tool_calls: RefCell::new(Vec::new()),
             compacted_up_to: RefCell::new(compacted_up_to),
             ancestor_prefix_len: Cell::new(ancestor_prefix_len),
         })
     }
 
-    pub async fn handle_content(
+    pub fn handle_content(
         &self,
         message: InfinityMessage,
         message_id: String,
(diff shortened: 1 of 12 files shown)


crates/infinity-agent-core/src/system/local/router.rs
```diff
@@ -249,11 +252,26 @@ async fn route_loop<C, S, H, O, F>(
...
+    // immediately. With per-thread tasks, both would be retained until the
+    // thread's next message, which adds up across many idle threads.
+    let mut drivers = FuturesUnordered::new();
 
     loop {
         let msg: Option<RoutedMessage<O::SubscribeRequest>> = tokio::select! {
             biased;
             _ = shutdown.cancelled() => None,
+            exited = drivers.next(), if !drivers.is_empty() => {
+                let exited: String = exited.expect("bug: drivers is non-empty");
```
MingweiSamuel:
```suggestion
                let exited: String = exited.expect("bug: drivers is empty");
```

crates/infinity-agent-core/src/system/local/router.rs
```diff
@@ -249,11 +252,26 @@ async fn route_loop<C, S, H, O, F>(
...
+                // Only remove the exited driver's own entry: if the thread
+                // already respawned, the new entry's channel is still open.
+                if workers.get(&exited).is_some_and(|w| w.input_tx.is_closed()) {
+                    workers.remove(&exited);
+                }
```
MingweiSamuel, line 272: ABA problem here, ideally would use a slotmap or something. but probably fine?

crates/infinity-agent-core/src/system/local/router.rs
```diff
@@ -249,11 +252,26 @@ async fn route_loop<C, S, H, O, F>(
 {
     let mut workers: HashMap<String, WorkerChannels<O::SubscribeRequest>> = HashMap::new();
```
MingweiSamuel, line 253: What are the string keys here?

crates/infinity-agent-core/src/event_processor.rs
```diff
@@ -281,8 +270,37 @@ impl<C: ConversationStore, S: StateStore> HistoryManager<C, S> {
...
+    /// must be stale.
+    fn match_tool_result(&self, result_id: &str) -> ToolResultMatch {
+        for msg in self.history.borrow().iter().rev() {
+            match msg {
+                InfinityMessage::ToolCall { call, .. } => {
+                    if call.id == result_id {
+                        // Its result would have been seen before the call in
+                        // a backwards walk, so this call is unanswered.
+                        return ToolResultMatch::Unanswered;
+                    }
```
MingweiSamuel, line 288: Nice, so the assumption is the tool call and (eventual) tool result will not ever be too far apart in history?

crates/infinity-agent-core/src/stores.rs
```diff
@@ -390,8 +392,7 @@ pub struct ThreadState {
 /// subscription tracking in process memory.
 #[derive(Clone, Default)]
 pub struct InMemoryStateStore {
-    #[expect(clippy::type_complexity, reason = "shared state")]
-    processed_ids: Arc<Mutex<HashMap<String, (HashSet<String>, HashSet<String>)>>>,
+    processed_ids: Arc<Mutex<HashMap<String, HashSet<String>>>>,
     metadata: Arc<Mutex<HashMap<String, serde_json::Value>>>,
     subscriptions: Arc<Mutex<HashMap<String, HashSet<String>>>>,
     pending_user_choices: Arc<Mutex<HashMap<String, Vec<UserChoice>>>>,
```
MingweiSamuel, line 394: we gotta get rid of these bare strings

MingweiSamuel (approved):
Nice

crates/infinity-agent-core/src/system/local/router.rs
shadaj: Thread IDs, we should probably newtype hmm.

crates/infinity-agent-core/src/event_processor.rs
shadaj: Yes because the only trailing tool calls / results will be for concurrent calls, of which there should not be that many.

PR #110, opened 2026-09-01
refactor(provider)!: rig-free provider stack, native Bedrock, minimal deps; refreshed scale claims

shadaj:
Remove `rig` from the core provider/agent stack. `infinity-provider-protocol` now owns a
minimal model API; the Bedrock provider talks to the official AWS SDK directly
(eliminating the maintained `rig-bedrock` patch); rig survives only as an optional
bridge crate. Re-measured the memory benchmark and refreshed the landing/README claims.

BREAKING CHANGE: `infinity-provider-protocol` types replace rig types throughout;
`ModelProvider::invoke_model` returns `ModelStream`; the `remote` module and mock are
behind the `remote`/`mock` features; the provider wire `Chunk` frame shape changed; the `rig-mock`
and `rig-bedrock-patched` crates are gone.


crates/infinity-agent-core/src/snapshots/infinity_agent_core__event_processor__tests__associative_subscription_event_inlined.snap
```diff
@@ -14,15 +14,12 @@ expression: hm.history.into_inner()
...
       "id": "tc-cmd",
-      "call_id": null,
       "function": {
         "name": "execute_command",
         "arguments": {
           "command": "make build"
         }
-      },
-      "signature": null,
-      "additional_params": null
```
akainth015, line 25: It's not clear to me why these changed, though the fields are `null` so we may not care.

crates/infinity-agent-core/src/system/test_support.rs
```diff
@@ -252,15 +260,18 @@ pub(crate) async fn wait_idle<Sub: Send + 'static>(running: &mut RunningSystem<S
...
-            if let rig::message::Message::User { content } = m
-                && let UserContent::ToolResult(r) = content.first()
-                && let rig::message::ToolResultContent::Text(t) = r.content.first()
+            if let infinity_provider_protocol::message::Message::User { content } = m
+                && let Some(UserContent::ToolResult(r)) = content.first()
+                && let Some(infinity_provider_protocol::message::ToolResultContent::Text(t)) =
+                    r.content.first()
             {
-                Some(t.text)
+                Some(t.text.clone())
```
akainth015, line 274: `ToolResultContent::Text(t)` is now a type that needs to be `.clone()`'d? `&str`?

crates/infinity-agent-core/src/system/tests.rs
```diff
@@ -587,13 +592,13 @@ async fn subscribe_mid_thinking_replays_current_thinking() {
             let (running, mut rx, mut ctrl, _conv) = start_system(vec![], None);
             running.send_user_text("t1", "think hard").await;
             let _req = ctrl.next_request().await;
-            ctrl.send_chunk(rig::streaming::RawStreamingChoice::ReasoningDelta {
+            ctrl.send_chunk(StreamChunk::ReasoningDelta {
                 id: None,
-                reasoning: "deep ".into(),
+                text: "deep ".into(),
```
akainth015, line 597: `ReasoningDelta` renames the `reasoning` field?

akainth015 (commented):
Partial review, I haven't yet finished because it's a massive diff 😅

crates/infinity-agent-core/src/snapshots/infinity_agent_core__event_processor__tests__associative_subscription_event_inlined.snap
shadaj: We just removed some extraneous fields that aren't relevant to us.

crates/infinity-agent-core/src/system/tests.rs
shadaj: Not a strictly necessary change, but it is safe because we only persist the final complete reasoning not the deltas.

crates/infinity-agent-core/src/system/test_support.rs
shadaj: The clone was always there, just hidden: rig's `OneOrMany::first()` returns an owned `T` by cloning it internally (`self.first.clone()`), so the old code cloned the whole `Text` struct on every call. With `Vec`, `.first()` returns `Option<&Text>`, so the clone becomes explicit at the use site — and only of the `String` field, and only where we actually need ownership (this helper collects `Vec<String>`). Read-only call sites now borrow with no clone at all.

crates/infinity-provider-protocol/Cargo.toml
```diff
@@ -5,22 +5,38 @@ edition = "2024"
...
+    "dep:uuid",
+    "dep:async-stream",
+    "dep:tracing",
+    "futures-util/sink",
+]
 
 [dev-dependencies]
-rig-mock = { path = "../rig-mock" }
+# Self-dependency so this crate's own tests cover the feature-gated modules.
+infinity-provider-protocol = { path = ".", features = ["mock", "remote"] }
```
akainth015, line 39: This is a cool trick!

crates/rig-mock/Cargo.toml
```diff
@@ -1,16 +1,20 @@
...
+name = "infinity-provider-rig"
 version = "0.1.0"
 edition = "2024"
 license = "Apache-2.0"
 
 [dependencies]
+infinity-provider-protocol = { path = "../infinity-provider-protocol" }
 rig-core = { workspace = true }
-serde = { workspace = true }
 serde_json = { workspace = true }
```
akainth015, line 10: It looks like you don't need `serde_json` either

akainth015 (approved):
Looks progressive to me 🙂 we get to use whatever TLS library we please

PR #111, opened 2026-09-01
ci: install jj via taiki-e/install-action instead of cargo install

shadaj:
Replace the `cargo install --locked --bin jj jj-cli` step in the test job of
.github/workflows/ci.yml with taiki-e/install-action (tool: jj-cli), pinned to
the full SHA of the v2 tag per the repo's action-pinning convention. jj-cli is
not in install-action's built-in manifests, so it is installed via the
cargo-binstall fallback, fetching prebuilt binaries instead of compiling from
source, which speeds up CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,7 +89,9 @@ jobs:
       - uses: ./.github/actions/use-sccache
 
       - name: Install jj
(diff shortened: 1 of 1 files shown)


akainth015 (approved):
What's the benefit

shadaj:
Zoom zoom CI

PR #114, opened 2026-09-03
feat(infinity-agent-core)!: allow spawning subagents with a fresh (no prefix) context

akainth015:
* Add a `fresh_context` boolean to the `spawn_thread` tool: when true, the child thread starts with an **empty** history and sees only the provided instructions, instead of inheriting the parent conversation prefix
* Introduce `SpawnContext` (`Inherit` / `InheritUpTo(n)` / `Fresh`) replacing the `spawn_order_override: Option<usize>` parameter of `ConversationStore::spawn_thread`
* Persist freshness as a thread attribute (`ThreadInfo::fresh_context`, new `fresh_context` column in the DSQL `thread_hierarchy` table) and add `ConversationStore::is_fresh_context_thread`
* `load_history_with_ancestors` now truncates the ancestor chain at the deepest fresh boundary, so neither a fresh thread nor its descendants can see context from before the fresh spawn (including grandparents)
* Seed fresh children with a synthetic `spawn_thread` tool call in their own store so the instruction tool-result pairs with a matching call
* Document fresh-context subagents in the default prompt

BREAKING CHANGE: `ConversationStore::spawn_thread` takes `SpawnContext` instead of `Option<usize>`, and implementations must provide `is_fresh_context_thread`.


crates/infinity-agent-core/src/traits.rs
```diff
@@ -6,6 +6,21 @@ use rap_protocol::{ChoiceId, ThreadId, ToolCallId};
 use crate::message::{InfinityMessage, InputMessage};
 use crate::system::UserChoice;
 
+/// How much of the parent's conversation history a newly spawned child
+/// thread inherits.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum SpawnContext {
+    /// Inherit the parent's full history up to the moment of the spawn.
+    Inherit,
```
shadaj, line 14: Does anyone use this other than tests?

crates/infinity-agent-core/src/system/tests.rs
```diff
@@ -1508,3 +1508,71 @@ async fn overflow_after_interrupt_drops_user_inputs_and_stops() {
         })
         .await;
 }
+
+/// A `fresh_context: true` spawn creates a child that inherits no parent
+/// history: its first model request contains only the synthetic spawn call
+/// and the self-contained instructions, none of the parent's messages.
+#[tokio::test(flavor = "current_thread")]
+async fn fresh_context_spawn_has_no_parent_history() {
```
shadaj, line 1516: We should also have a test for doubly nested thread, `t1 -> spawn no context t2 -> spawn inherit t3`, `t3` should only see `t2` context.

crates/infinity-agent-core/src/traits.rs
akainth015: Yes! It's used [here](https://github.com/hydro-project/infinity/pull/114/changes#diff-cadd610d94a0d7864564294f64c9568c39f127b678d1f9b105e4dcaf84e23247R106) in the `SpawnThreadTool`.

crates/infinity-agent-core/src/traits.rs
```diff
@@ -55,7 +70,23 @@ pub trait ConversationStore: Send + Sync + Clone {
...
 
-        let ancestors = self.get_ancestor_chain(thread_id).await?;
+        let mut ancestors = self.get_ancestor_chain(thread_id).await?;
+
+        // Fresh-context threads sever history inheritance: nothing from
+        // before the fresh spawn is visible. Find the deepest fresh boundary
+        // in the chain (the leaf itself, or an ancestor) and drop every
+        // ancestor slice above it. Entry `i` holds ancestor `i`'s own slice,
+        // so when ancestor `i` is the fresh thread its slice is kept.
+        if self.is_fresh_context_thread(thread_id).await? {
```
shadaj, line 80: Let's create an issue for simplifying + cleaning up this code by having the ancestor chain itself stop at a fresh context thread.

shadaj (approved):
Nice!

PR #121, opened 2026-09-04
feat(infinity-provider-protocol,infinity-provider-bedrock,infinity-agent-core)!: provider-classified errors, provider-owned timeouts, and oversized-input recovery

shadaj:
## `infinity-provider-protocol`: structured, classified errors

* New `ErrorClass` enum (`Transient` / `Throttled` / `ContextOverflow` /
  `Fatal`) — providers declare how callers should react; the agent runtime
  never parses error message strings.
* `CompletionError::ProviderError` now carries `{ message, class }`
  (breaking); added `CompletionError::provider(class, msg)` and
  `CompletionError::class()` (RequestError/JsonError → Fatal,
  ResponseError → Transient).
* Remote wire protocol forwards errors as `WireError { message, class }` so
  classification survives the Unix-socket boundary (`Fatal` default for
  older providers). Covered by `error_class_survives_the_wire`.

## `infinity-provider-bedrock`: timeouts + classification live here now

* 60s request-initiation timeout around `ConverseStream` (Bedrock
  occasionally black-holes requests), reported as `Transient` so core
  retries. Deliberately **no** mid-stream/inactivity timeout anywhere: once
  a stream is live it is never artificially cut off by us. Tested with a
  never-responding HTTP connector under paused time.
* `classify_bedrock_error` maps service codes + message heuristics (moved
  out of core) to `ErrorClass`: throttling → `Throttled`, input-too-long /
  context-limit → `ContextOverflow`, backend hiccups → `Transient`, rest
  `Fatal`. Mid-stream recv errors are classified the same way.
* New non-default `live-tests` feature (`tests/live.rs`) hitting real
  Bedrock with local credentials — overflow classification, unknown-model
  fatality, happy path — instead of tautological string-table unit tests.
  All pass (and caught a bad token estimate during development).
* `infinity-provider-rig` classifies rig errors from common
  OpenAI-compatible phrasings.

## `infinity-agent-core`: no timeouts, class-driven retries, oversized-input fix

`run_completion` drops the 60s initiation timeout and the 120s mid-stream
stall timer; retries are driven purely by `ErrorClass` (Throttled → 30s
backoff, Transient → short retry, Fatal → give up), capped at
`MAX_COMPLETION_RETRIES`.

**Three-phase pending items** (fixes: an oversized user input / tool result
used to be persisted and permanently wedge the thread):

* `HistoryManager` inputs move through **not-known-safe**
  (`unvalidated_items`) → **known-safe** (`pending_items`, promoted the
  moment the model streams any output — proof the context didn't overflow;
  model-produced content goes here directly) → **synced**. `sync()` only
  persists known-safe items; a sequentiality assert makes committing model
  output over unvalidated inputs impossible.
* Interrupts: cancelling before any model output keeps the inputs in memory
  (re-sent with the interrupting message next round) but persists nothing;
  after output, input + partial output persist as before.
* `safe_spawn_point()` stays before unvalidated inputs (children inherit
  history from the store).

**Overflow recovery** (`recover_from_overflow`):

* Exactly one unvalidated item and it's replaceable (tool result /
  subscription event): the culprit is unambiguous → replace its body with
  `TOOL_RESULT_TOO_LARGE_PLACEHOLDER`, retry. (The replace helper skips
  bodies already equal to the placeholder, so a second overflow falls
  through to drop mode instead of retrying forever.)
* Anything else (user text present, multiple items, or the retry overflowed
  again): settle replaceables with the shared `TOOL_CALL_INTERRUPTED_TEXT`
  (a "too large" note could mislead the model into re-running the tool),
  drop the user inputs (`drop_unvalidated_user_inputs`, dedup IDs
  forgotten), and **stop** — dropping the user's words is never followed by
  a silent retry. Tool calls always stay answered in memory; nothing
  unvalidated ever persists. TODO left for blocking on a lower-threshold
  compaction when committed history itself is too close to the limit.

## Tests (each behavior landed red-first)

* event_processor: oversized user input dropped & unpersisted; oversized
  tool result / subscription event → placeholder retry; tool result + user
  input → interrupted-settle + drop + **no retry**; second overflow gives up
  keeping the call answered; interrupt before/after model output;
  throttled/transient retry vs fatal no-retry; `sync()` skips unvalidated;
  flush-with-unvalidated panics; spawn-point exclusion.
* system: interrupt-before-output persists nothing until validated; overflow
  after interrupt drops both inputs and stops (thread recovers on next
  message); oversized tool result never wedges the thread end-to-end;
  shutdown persists validated but not unvalidated tool results; rebooted
  session settles a persisted unanswered tool call via the synthetic
  "interrupted" result.

`./check.bash`: fmt, clippy, and full workspace tests pass (99 core tests +
all other crates); the web-e2e step stalled on a sandboxed `npm ci`
(network), unrelated to these changes.


crates/infinity-agent-core/src/event_processor.rs
```diff
@@ -1148,6 +1312,76 @@ where
...
+    /// replaced with [`TOOL_RESULT_TOO_LARGE_PLACEHOLDER`] and the
+    /// completion should be retried.
+    RetryWithPlaceholder,
+    /// Anything else: replaceable items were settled with
+    /// [`TOOL_CALL_INTERRUPTED_TEXT`], user inputs were dropped (count
+    /// `usize`, they have no safe substitute), and the completion must
+    /// fail. Dropping is never followed by a retry — the user clearly
+    /// wanted to say something, and silently re-running the round without
+    /// their words would act on stale instructions.
+    DroppedInputs(usize),
```
MingweiSamuel, line 1354: What is the `usize`? Number of inputs droped?

crates/infinity-agent-core/src/event_processor.rs
```diff
@@ -1148,6 +1312,76 @@ where
...
+fn recover_from_overflow<C: ConversationStore, S: StateStore>(
+    history: &HistoryManager<C, S>,
+) -> OverflowRecovery {
+    if history.unvalidated_len() == 1
+        && history.replace_unvalidated_tool_results(TOOL_RESULT_TOO_LARGE_PLACEHOLDER)
+    {
+        return OverflowRecovery::RetryWithPlaceholder;
+    }
+    history.replace_unvalidated_tool_results(TOOL_CALL_INTERRUPTED_TEXT);
+    OverflowRecovery::DroppedInputs(history.drop_unvalidated_user_inputs())
```
MingweiSamuel, line 1382: Is `replace_unvalidated_tool_results` ever *not* called with `TOOL_CALL_INTERRUPTED_TEXT`?

crates/infinity-agent-core/src/event_processor.rs
shadaj: Yep!

crates/infinity-agent-core/src/event_processor.rs
shadaj: Right above! If there is a single source of overflow then we give the agent a richer message.

MingweiSamuel (approved):
Seems good

PR #123, opened 2026-09-15
refactor(infinity-provider-bedrock): collapse nested if into match guard

akainth015:
## Summary

Fixes the clippy 1.95 `collapsible_match` lint (denied via `-D warnings`) in the Bedrock stream delta handling by folding the `tool_call.is_none()` check into a match guard, as clippy suggests.

No behavior change: text deltas arriving during a tool call now fall through to the no-op `_ => {}` catch-all arm instead of being dropped inside the arm body.

## Testing

- `cargo clippy -p infinity-provider-bedrock --all-targets -- -D warnings` passes
- `cargo test -p infinity-provider-bedrock` — 18 tests pass
- Full `./check.bash` passes locally (docs formatting and Playwright e2e steps skipped for missing local prerequisites)

## Note

`feat/fresh-subagents` carries an equivalent fix (differing only in comment placement and an explicit empty `Text(_)` arm); whichever lands first makes the other redundant.


akainth015:
Closing: this was a clippy 1.95 false positive. On clippy 1.98.1 (current stable, which CI runners track), `main` passes `cargo clippy --all-targets -- -D warnings` unchanged, so no code change is needed. The equivalent commit on `feat/fresh-subagents` can also be dropped.

PR #124, opened 2026-09-17
fix(infinity-provider-bedrock): classify "unexpected error during processing" as transient

shadaj:
The Bedrock error message "The system encountered an unexpected error
during processing. Try your request again." was falling through to
ErrorClass::Fatal because the existing heuristic in
classify_bedrock_error only matched the older wording "unexpected error
when processing the request". When the error arrives without a usable
service error code (e.g. a mid-stream event error), the message
heuristics are the only signal, so the request was never retried.

Add message heuristics for "unexpected error during processing" and
"try your request again" to the Transient branch.


akainth015 (approved):
Would be nice to have a canonical retry mechanism for transient errors.

PR #126, opened 2026-09-18
fix(infinity-daemon): tolerate high latency when connecting to remote daemons

shadaj:


akainth015:
What's the use case here?

shadaj:
SSH-ing from planes, where establishing the tunnel can take some time.

akainth015 (approved):
We live in a society

PR #127, opened 2026-09-18
fix(infinity-provider-bedrock): enable fine-grained tool streaming to prevent mid-tool-call stream timeouts

shadaj:
Root cause of the "stream error timeout" during large tool calls: without
fine-grained tool streaming, the Anthropic backend buffers and validates
each tool-input parameter server-side before streaming it back. A large
tool call therefore produces a long silent gap right after the tool-use
block starts (the point where we emit the first ToolCallDelta), and
Bedrock's stream watchdog kills the idle stream with a
ModelTimeoutException. No client-side timeout is involved — the audit
confirmed core, the provider socket, and the AWS SDK (event-stream ops are
exempt from stalled-stream protection) apply none mid-stream.

Changes:

* lib.rs: add the `fine-grained-tool-streaming-2025-05-14` anthropic beta
  flag to every default model's additional_request_params (the 1m alias
  keeps both betas; flags are now named constants). The modern per-tool
  `eager_input_streaming` field that replaces this header is not
  expressible on Bedrock: the Converse ToolSpecification does not model it
  even in the latest aws-sdk-bedrockruntime (verified against 1.145.0),
  and additionalModelRequestFields cannot attach per-tool attributes. The
  comment records the migration path (switch to the per-tool field in
  convert::tool_config once the SDK exposes it). The header's only
  documented incompatibility (computer-use/browser-use system toolsets)
  cannot occur since the converter only emits Tool::ToolSpec.

* stream.rs: fine-grained streaming skips server-side JSON validation, so
  a tool input that fails to parse at ContentBlockStop is now classified
  ErrorClass::Transient (retry regenerates the turn) instead of the fatal
  JsonError. Genuine max-tokens truncation still surfaces as its own fatal
  error at MessageStop.

* Tests: assert the beta flag on the default and 1m-alias params, add a
  test that every default model carries the flag, and update the malformed
  tool-input test to expect a transient classification.


crates/infinity-provider-bedrock/src/lib.rs
```diff
@@ -28,6 +28,29 @@ type BoxError = Box<dyn std::error::Error + Send + Sync>;
...
+/// before streaming it back, so a large tool call produces a long silent
+/// gap after the tool-use block starts — long enough for Bedrock's stream
+/// watchdog to kill the request with a `ModelTimeoutException`. With the
+/// flag, input fragments stream as they are generated and the stream never
+/// goes idle. The modern per-tool `eager_input_streaming` field replaces
+/// this header on the Anthropic-native API, but Bedrock's Converse
+/// `ToolSpecification` does not model it (checked through
+/// aws-sdk-bedrockruntime 1.145.0 — only `strict` has been added), so the
+/// legacy header is the only Bedrock-expressible mechanism; switch to the
+/// per-tool field in `convert::tool_config` once the SDK exposes it. The
```
akainth015, line 42: Create a GitHub issue to track this?

crates/infinity-provider-bedrock/src/lib.rs
```diff
@@ -28,6 +28,29 @@ type BoxError = Box<dyn std::error::Error + Send + Sync>;
...
+/// header is incompatible only with computer-use/browser-use system
+/// toolsets, which this provider never emits.
+///
+/// Trade-off: streamed fragments are not validated server-side, so the
+/// accumulated tool input may not parse — see `finish_tool_call` in
+/// `stream.rs` for how that is classified.
+const FINE_GRAINED_TOOL_STREAMING_BETA: &str = "fine-grained-tool-streaming-2025-05-14";
+
+/// Anthropic beta flag enabling the 1M-token context window.
+const CONTEXT_1M_BETA: &str = "context-1m-2025-08-07";
```
akainth015, line 52: Is this necessary? Why now?

crates/infinity-provider-bedrock/src/lib.rs
```diff
@@ -173,12 +196,14 @@ fn default_models() -> Vec<BedrockModel> {
         "thinking": {
             "type": "adaptive",
             "display": "summarized"
-        }
+        },
+        "anthropic_beta": [FINE_GRAINED_TOOL_STREAMING_BETA]
```
akainth015, line 200: What would happen if `anthropic_beta` is passed to e.g. OpenAI's Astra? Would it be ignored or cause an error?

akainth015 (approved):
The gist seems fine. Might require some small refactoring to add non-Anthropic models for Bedrock.

crates/infinity-provider-bedrock/src/lib.rs
shadaj: This is applied as config per-model, so wouldn't be applied.

crates/infinity-provider-bedrock/src/lib.rs
shadaj: Was already there, just extracted as a const.

crates/infinity-provider-bedrock/src/lib.rs
shadaj: #128

